PR #66, opened 2021-02-16
Fetch executive proposals from Github index rather than CMS

jparklev:


lib/api.ts
```diff
@@ -4,58 +4,55 @@ import invariant from 'tiny-invariant';
...
+
+    invariant(content && title && summary && address && date, 'Invalid proposal document');
+    proposals.push({
+      about: content,
+      title,
+      proposalBlurb: summary,
+      key: slugify(title),
+      address: address,
+      date: String(date),
+      active: proposalIndex[network] ? proposalIndex[network].includes(proposalLink) : true
```
jparklev, line 50: so would the `else` case here mainly trigger when we're on a testchain?

tyler17: yeah exactly

PR #77, opened 2021-06-16
fix: Update DSR amount, handle temp files for windows users

rafinskipg:
- DSR is a bignumber, had to multiply by 100 because order of magnited might be wrong since we are getting:
0.000100000000000293160873512193275890079183337856595266653047766296648530270031623894233239896534063

- tmp directory handling for windows users


b-pmcg:
the DSR is 0.01% you can see from here: https://daistats.com/#/ecosystem

adamgoth (approved):

PR #79, opened 2021-06-17
feat: Add delegates page

rafinskipg:
### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/110/create-delegates-page

### What does this PR do?

Adds delegates page

### Steps for testing:

1. Go to `/delegates`
2. Should match figma


components/Icon/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+      </g>
+    ),
+    viewBox: '0 0 15 15',
+  },
+};
+
+export default function Icon({
```
b-pmcg, line 30: @rafinskipg this component can be imported from the `dai-ui-icons` package directly, any reason you needed to export it here?

For ex: https://github.com/makerdao/governance-portal-v2/blob/da2e66d0397ce09553fe1ce284674bb2b6c2e690/components/Header.tsx#L5

rafinskipg: hey @b-pmcg yes, my plan was to migrate the icons i am adding here to dai-ui when we have permissions to release new versions. This Icon component is a direct copy of the code in dai-ui

Is there any way to add new icons to dai-ui right now?

PR #80, opened 2021-06-17
fix: Absolute imports

rafinskipg:
### What does this PR do?

add @lib @components @stores imports

include types as type definitions , and add standard unified way of importing them


tsconfig.json
```diff
@@ -13,7 +13,16 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "baseUrl": "./",
+    "paths": {
+      "@lib/*": ["lib/*"],
+      "@components/*": ["components/*"],
+      "@stores/*": ["stores/*"],
```
adamgoth, line 21: With `baseURL` specified to `./`, these shortcuts are actually the same as just using the actual path. `@lib/*` does the same thing as `lib/*`. So I'm not sure we need the shortcut with the `@`. Thoughts?

rafinskipg: have you tried it without the '@' ? i don't know if it works

KenG-pro (approved):

KenG-pro (approved):

tsconfig.json
adamgoth: Using `lib/* works the same as `@lib/*`. I just made sure. Adding `"baseUrl": "./"` in `tsconfig.json` makes it work.

rafinskipg: ok, so we replace all the @ stuff

tyler17 (approved):

PR #82, opened 2021-06-23
some refactor

rafinskipg:
### What does this PR do?
Changed process.env variables in a single file called config. This way it can be mocked in tests or changed easily. All the process env vars are defined there.

Added global typings for things in window 

Changed some minor typescript errors


adamgoth (approved):
You can delete the non-relevant sections from the PR template but I think it's good to include a description of what the PR does in case we need to go back and look at it at some point 😅

That said, refactor looks good to me. Curious if @tyler17 or @b-pmcg have any thoughts though.

tyler17 (approved):

PR #89, opened 2021-07-07
Delegate executive voting

adamgoth:
Updates how vote delegate is used in accounts store

Checks for vote delegate existence in vote modal and uses vote delegate to vote on executive if it exists

You will need the latest dai.js changes on `vote-delegation` branch to test this locally.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -295,3 +295,8 @@ export const sortBytesArray = _array =>
 
 export const formatRound = (num, decimals = 2) =>
   isNaN(num) ? '----' : round(num, decimals).toLocaleString({}, { minimumFractionDigits: decimals });
+
+export const getVotingWeightCopy = (isVotingDelegate: boolean): string =>
+  isVotingDelegate
+    ? 'Your voting weight is made up of the MKR delegated to your delegate contract.'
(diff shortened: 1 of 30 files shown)


lib/utils.ts
```diff
@@ -295,3 +295,8 @@ export const sortBytesArray = _array =>
 
 export const formatRound = (num, decimals = 2) =>
   isNaN(num) ? '----' : round(num, decimals).toLocaleString({}, { minimumFractionDigits: decimals });
+
+export const getVotingWeightCopy = (isVotingDelegate: boolean): string =>
```
rafinskipg, line 299: Instead of separating it into a function that is on a generic folder "utils" i would create a simple component under the relevant module "voting" or whatever.

Utils is a concept that we should discontinue, because leads to mix of responsabilities mixed

rafinskipg (changes requested):

lib/utils.ts
adamgoth: Fair enough, moved the function to `lib/polling/`

adamgoth:
Merging this into the feature branch because everything looks to be working on the branch URL and the feature branch needs thorough code reviews anyways

PR #90, opened 2021-07-07
Fix/refactor mkr input

rafinskipg:
### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/128/update-mkrinput-component

### What does this PR do?

Changes the MKR input to be reused in all the places we input MKR

Removes the use of Currency in favor of BigNumber

Refactor's some calls to reduce code duplicity.

### Steps for testing:

1. Go to http://localhost:3000/executive?network=kovan and do a deposit and withdrawal and test that it works
2. Test that deposits for delegate / undelegate still work

image


pages/index.tsx
```diff
@@ -329,18 +328,25 @@ export default function Index({
...
-    }
-  };
+  try {
+    const [proposals, polls, blogPosts] = await Promise.all([
+      getExecutiveProposals(),
+      getPolls(),
+      getPostsAndPhotos()
+    ]);
+
+    console.log('EEEE');
```
adamgoth, line 339: EEEEEEEEEE 🙀

pages/executive.tsx
```diff
@@ -208,8 +207,8 @@ export const ExecutiveOverview = ({ proposals }: { proposals: Proposal[] }): JSX
               >
                 <Text sx={{ py: 2 }}>
                   An executive vote has passed to update the Chief to a new version. You have{' '}
-                  <b>{lockedMkrOldChief.toBigNumber().toFormat(lockedMkrOldChief.gte(0.01) ? 2 : 6)} MKR</b>{' '}
```
adamgoth, line 211: I don't think you want to remove`toBigNumber` here since we're getting a currency object back.

adamgoth (commented):
I'm not sure if it was a previous issue or if it was introduced in this PR but I'm unable to type a `0` after a decimal in the put. For example, try to type in `1.05`.

rafinskipg:
@adamgoth  Yes, it is the default behaviour of HTML number input and it depends on your locale settings. You can input 1,05  or 1.05 depending on your browser settings

Before we were doing some weird handling and almost treating input type number as a text.

adamgoth:
> @adamgoth Yes, it is the default behaviour of HTML number input and it depends on your locale settings. You can input 1,05 or 1.05 depending on your browser settings
> 
> Before we were doing some weird handling and almost treating input type number as a text.
> 
> I know we should allow 1,05 and 1.05 but i think we need to find a better solution for handling those and still dealing with input type number

Hmm, to me we should expect it to be inputted the same way as it's displayed as the MKR balance below the input. But as it is, I'm unable to type either `1,05` or `1.05`.

components/MKRInput.tsx
```diff
@@ -119,6 +123,11 @@ export function MKRInput({
           Please enter a valid amount.
         </Text>
       )}
+      {errorInvalidFormat && (
+        <Text sx={{ color: 'error', fontSize: 2 }} data-testid="mkr-input-error">
+          Please, enter a valid number.
```
adamgoth, line 126: No `,` necessary here.

adamgoth (approved):
Looks good 🤘  just left a grammar-related comment on the error message.

components/MKRInput.tsx
rafinskipg: Why, not? I , like, commas, thanks, see, you, later

PR #95, opened 2021-07-09
fix: Add ENS, fixes tests ( A bit )

rafinskipg:
### Link to Clubhouse story

Add ENS name resolution.

Add some UI status when clicking on wallet connect.

Fixes tests so they can run.

Removes package-lock.json, leaving only yarn.lock, to avoid conflicts in versions.


components/es/burnModal/MKRAmount.tsx
```diff
@@ -34,11 +34,6 @@ const MKRAmountView = ({ setBurnAmount, burnAmount, mkrBalance }: Props) => {
           value={burnAmount.toBigNumber()}
         />
       </Box>
-
-      <Flex mt={3} sx={{ alignItems: 'center', justifyContent: 'flex-start', width: '100%' }}>
-        <Text variant="caps">MKR Balance In Wallet</Text>
-        <Text ml={3}>{mkrBalance ? mkrBalance.toString() : '---'}</Text>
-      </Flex>
     </>
```
adamgoth, line 37: I'm not sure what this was doing but it's safe to delete?

pages/esmodule.tsx
```diff
@@ -106,7 +83,10 @@ const ESModule = () => {
             typeof totalStaked !== 'undefined'
               ? canFire
                 ? 100
-                : totalStaked.mul(100).div(thresholdAmount).toFixed()
+                : totalStaked
+                    .mul(100)
+                    .div(thresholdAmount || 1)
```
adamgoth, line 88: Are you sure adding these updates around `thresholdAmount` don't change any of the intended logic?

pages/esmodule.tsx
```diff
@@ -154,7 +134,7 @@ const ESModule = () => {
           )}
         </DialogContent>
       </DialogOverlay>
-      {cageTime && cageTime.gt(0) && (
+      {cageTime && (
```
adamgoth, line 137: Why remove `cageTime.gt(0)`?

components/es/burnModal/MKRAmount.tsx
rafinskipg: yes, was duplicated because the balance is in the MKRInput

pages/esmodule.tsx
rafinskipg: No, i'm not, but i guessed that if thresholdAmount is not defined and we divide by 0 it will give an error.. But probably we need to check

pages/esmodule.tsx
rafinskipg: It seems that cageTime is a Date object, i think this was a bug. 
https://github.com/makerdao/dai.js/blob/0e679826a841c654ee9c49c81101e6092b486753/packages/dai-plugin-mcd/src/schemas/end.js#L15

It also seems that the format date used below receives a Date object. I think someone has mistaken it by a CurrencyObject or a bignumber. 

But i'm not sure, i don't know how to activate the cageTime

pages/esmodule.tsx
adamgoth: I see. Maybe it's better to not render the progress ring at all if `thresholdAmount` is undefined. Progress ring would be kind of useless if we don't know the total progress to be made.

rafinskipg: makes sense

rafinskipg: At the end i checked if there is no thresholdAmount, show 0

adamgoth (approved):

PR #102, opened 2021-07-14
refactor delegations fetching

rafinskipg:
### What does this PR do?

- Fetches kovan delegates from https://github.com/makerdao-dux/voting-delegates
- Fetches kovan and mainnet delegates from serverside paths. Disabling clientside fetching and minimizing the number of requests.
- Remove last voted date 
- ⚠️ Introduces `getMaker(network)` that returns a instance of maker for the specified network. This is usefull to fetch server side kovan information
- Adds styles for tables for markdown rendering as suggested by @LongForWisdom
- Separate FScache into a module, add expiration date
- Fetch delegates through API requests on client side (with the cache included)

### Steps for testing:

1. Go to delegates page in kovan
2. See a recognized delegate and can access to their profile

1. See that http://localhost:3000/api/delegates?network=kovan works
2. See that http://localhost:3000/api/delegates/0xc8829647c8e4131a01354ccac993388568d12d00?network=kovan works
3. See that cache is being regenerated/dumped for delegates API every 30 seconds

image

image

image

### DIFF
--- a/components/delegations/DelegateCard.tsx
+++ b/components/delegations/DelegateCard.tsx
@@ -83,7 +83,7 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
 
             {!showLinkToDetail && (
               <Box>
-                <DelegateLastVoted delegate={delegate} />
+                {/*<DelegateLastVoted delegate={delegate} /> */}
                 <DelegateContractExpiration delegate={delegate} />
               </Box>
             )}
--- a/components/delegations/DelegateCard.tsx
+++ b/components/delegations/DelegateCard.tsx
@@ -23,6 +23,7 @@ type PropTypes = {
 };
 
 export default function DelegateCard({ delegate }: PropTypes): React.ReactElement {
+  const network = getNetwork();
   const bpi = useBreakpointIndex();
   const [showDelegateModal, setShowDelegateModal] = useState(false);
   const [showUndelegateModal, setShowUndelegateModal] = useState(false);
@@ -66,7 +67,12 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
 
           <Box sx={{ mt: 3 }}>
             {showLinkToDetail && (
-              <Link href={`/delegates/${delegate.voteDelegateAddress}`}>
+              <Link
+                href={{
+                  pathname: `/delegates/${network}/${delegate.voteDelegateAddress}`,
+                  query: { network }
+                }}
+              >
                 <a title="Profile details">
                   <Button sx={{ borderColor: 'text', width: '169px', color: 'text' }} variant="outline">
                     View Profile Details
--- a/components/delegations/DelegateCard.tsx
+++ b/components/delegations/DelegateCard.tsx
@@ -69,7 +69,7 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
             {showLinkToDetail && (
               <Link
                 href={{
-                  pathname: `/delegates/${network}/${delegate.voteDelegateAddress}`,
+                  pathname: `/delegates/${delegate.voteDelegateAddress}`,
                   query: { network }
                 }}
               >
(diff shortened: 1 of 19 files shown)


components/delegations/DelegateCard.tsx
```diff
@@ -66,7 +67,12 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
 
           <Box sx={{ mt: 3 }}>
             {showLinkToDetail && (
-              <Link href={`/delegates/${delegate.voteDelegateAddress}`}>
+              <Link
+                href={{
+                  pathname: `/delegates/${network}/${delegate.voteDelegateAddress}`,
+                  query: { network }
```
rafinskipg, line 73: delegates/network/ path was introduced to allow server side rendering with static properties

adamgoth: I'm curious to hear what others think but having network in the path feels clunky and somewhat unnecessary to me. As the devs here we use kovan a fair amount but I think 99% of our users will be on mainnet. We shouldn't optimize for using testnets. I think we should keep the network out of the path name and just use the query param. 

If we have to refetch data client side because we're on a testnet, I don't think it's a big deal. We are some of the only users who are using kovan in this case.

lib/api.ts
```diff
@@ -2,21 +2,19 @@ import uniqBy from 'lodash/uniqBy';
...
 import { CMSProposal } from 'types/proposal';
 import { BlogPost } from 'types/blogPost';
-import { DelegateContractInformation } from 'types/delegate';
 import { slugify } from '../lib/utils';
 import { parsePollMetadata } from './polling/parser';
 import { fetchGitHubPage } from './github';
-import fs from 'fs';
 import { config } from './config';
 import mockProposals from '../mocks/proposals.json';
+import { fsCacheGet, fsCacheSet } from './fscache';
```
adamgoth, line 17: 👍

pages/delegates/[network]/[address].tsx
```diff
@@ -4,18 +4,19 @@ import ErrorPage from 'next/error';
...
-import PrimaryLayout from '../../components/layouts/Primary';
-import SidebarLayout from '../../components/layouts/Sidebar';
-import Stack from '../../components/layouts/Stack';
-import SystemStatsSidebar from '../../components/SystemStatsSidebar';
-import ResourceBox from '../../components/ResourceBox';
+import PrimaryLayout from '../../../components/layouts/Primary';
+import SidebarLayout from '../../../components/layouts/Sidebar';
+import Stack from '../../../components/layouts/Stack';
+import SystemStatsSidebar from '../../../components/SystemStatsSidebar';
+import ResourceBox from '../../../components/ResourceBox';
```
Comment (adamgoth) [inline pages/delegates/[network]/[address].tsx:11]: You could just remove the `../../../` on these.

pages/delegates/[network]/[address].tsx
```diff
@@ -83,11 +84,13 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
 };
 
 export const getStaticPaths: GetStaticPaths = async () => {
-  const delegates = await fetchDelegates();
-  const paths = delegates.map(d => `/delegates/${d.voteDelegateAddress}`);
+  const delegatesMainnet = await fetchDelegates(SupportedNetworks.MAINNET);
+  const delegatesKovan = await fetchDelegates(SupportedNetworks.KOVAN);
```
Comment (adamgoth) [inline pages/delegates/[network]/[address].tsx:88]: Kind of the same comment I made above but this feels like a waste of a call, I don't think many people (besides us) will be using kovan very often in production.

adamgoth (commented):
Overall looks good to me, left some thoughts on using the network in the path. Curious to hear what @b-pmcg and @tyler17 think there.

components/delegations/DelegateCard.tsx
b-pmcg: Yeah I would prefer not to have the network in the path. Changes to the URL might be something we would want to run by governance also.

pages/delegates/[network]/[address].tsx
Comment (rafinskipg) [reply pages/delegates/[network]/[address].tsx:88]: We must fetch both if we want to use getStaticPaths and getStaticProps. If we don't want paths and props we can fetch only on demand on the client side.

PR #105, opened 2021-07-16
Remove time from executive posted date

adamgoth:
### What does this PR do?

Fixes issue that was brought up in Rocket Chat (confirmed desired fix with LFW):

Removes timestamp from executive posted date on overview card (instead of showing `0:00 UTC` for all)

### Steps for testing:

1. Go to `/executive`
2. See that the posted date on each card no longer shows the timestamp

Before:
Screen Shot 2021-07-16 at 1 23 26 PM

After:
Screen Shot 2021-07-16 at 1 23 56 PM

### DIFF
--- a/components/executive/ExecutiveOverviewCard.tsx
+++ b/components/executive/ExecutiveOverviewCard.tsx
@@ -9,7 +9,7 @@ import Bignumber from 'bignumber.js';
 
 import { Proposal } from 'types/proposal';
 import getMaker, { getNetwork } from 'lib/maker';
-import { formatDateWithTime } from 'lib/utils';
+import { formatDateWithTime, formatDateWithoutTime } from 'lib/utils';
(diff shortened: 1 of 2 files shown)


b-pmcg:
LGTM. Was it the dates that are wrong, or do they just not like seeing `0:00`?

b-pmcg (approved):

b-pmcg:
It looks like your test failed because of an async hangup. I re-ran it and it passed.

tyler17 (approved):

adamgoth:
> LGTM. Was it the dates that are wrong, or do they just not like seeing `0:00`?

The time is always `0:00` in the markdown but it's not accurate. It led to someone expecting the executive to actually be posted at that time but it's usually much later in the day than `0:00` UTC. So this just removes the confusion.

PR #106, opened 2021-07-16
Responsive delegate card

adamgoth:
Finishes adding responsive layout for delegate card

Also standardizes imports/exports for delegate components

See gif (note button alignment in gif was updated too):

https://user-images.githubusercontent.com/5225766/126156863-63df5224-a9f8-4044-8c9b-8dffa6a91a0b.mp4


components/delegations/DelegateCard.tsx
```diff
@@ -77,76 +79,74 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
...
+            </Flex>
 
-            <Box sx={{ mr: [0, 0, 4] }}>
+            <Box sx={{ mr: [0, 0, 4, 0, 4] }}>
               <Box sx={{ mb: 3 }}>
                 <Text as="p" variant="microHeading" sx={{ fontSize: [3, 5] }}>
                   {mkrStaked ? mkrStaked.toBigNumber().toFormat(2) : '0.00'}
                 </Text>
-                <Text as="p" variant="secondary" color="onSecondary">
+                <Text as="p" variant="secondary" color="onSecondary" sx={{ fontSize: ['13px', '15px'] }}>
```
b-pmcg, line 127: I think we should try and stick to the scaling system from the theme, so using `fontSize: [1, 2]` here would get us 12px and 14px respectively. 

If the designs call for 13px & 15px we should discuss why. If it's absolutely necessary to target those sizes we should align the theme spec with the design (or vice versa).

components/delegations/DelegateCard.tsx
```diff
@@ -77,76 +79,74 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
...
                   MKR delegated by you
                 </Text>
               </Box>
               <Box>
                 <Text as="p" variant="microHeading" sx={{ fontSize: [3, 5], color: 'secondaryMuted' }}>
                   Untracked
                 </Text>
-                <Text as="p" variant="secondary" color="onSecondary">
+                <Text as="p" variant="secondary" color="onSecondary" sx={{ fontSize: ['13px', '15px'] }}>
```
b-pmcg, line 135: same as above

components/delegations/DelegateCard.tsx
```diff
@@ -77,76 +79,74 @@ export default function DelegateCard({ delegate }: PropTypes): React.ReactElemen
...
           </Grid>
 
-          <Box
+          <Flex
             sx={{
-              textAlign: 'right',
-              width: ['100%', '100%', 'auto'],
-              display: ['flex', 'flex', 'block'],
+              width: ['100%', '100%', 'auto', '100%', 'auto'],
+              flexDirection: ['row', 'row', 'column', 'row', 'column'],
```
b-pmcg, line 145: Do we really have 5 breakpoints to account for? That seems excessive, maybe we could simplify and target 3 (mobile, tablet, desktop)?

components/delegations/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export * from './modals/InputDelegateMkr';
+export * from './modals/TransactionInProgress';
+export * from './modals/TxDisplay';
+export * from './modals/TxFinal';
+export * from './modals/UndelegateModal';
+export * from './DelegateCard';
+export * from './DelegateContractExpiration';
+export * from './DelegateDetail';
+export * from './DelegateLastVoted';
+export * from './DelegatePicture';
```
b-pmcg, line 13: Not totally opposed to this, but it's not my favorite :P Sometimes I find it makes debugging harder. But if we want to adopt this style maybe we should add it to our code standards file to try to achieve uniformity across our apps.

b-pmcg (approved):
Looks good to me, only request is to use the fontSizes scale instead of pixels if possible.

b-pmcg:
Oops, just realized you added the WIP tag, so I dismissed my approval, will wait until its ready.

components/delegations/DelegateCard.tsx
rafinskipg: Yeah. I don't even think tablet is meaningful these days anymore. If we optimize for phone and desktop we are covering already the most important things.
image

But this is something we can cover on another conversation

components/delegations/DelegateCard.tsx
rafinskipg: @tiago-varandas calling you here in the conversation.

tiago-varandas: Stick to the scaling system from the theme would make sense.

adamgoth: Agreed, updated back to 14 and 16px here (2, 3 on the scale)

components/delegations/DelegateCard.tsx
adamgoth: The breakpoints are part of the theme UI spec. So we could discuss if want to change this going forward.  

The challenge here is the sidebar layout. The width of the workable space on the page grows, then shrinks again when the sidebar appears, then continues to grow again.

components/delegations/index.ts
adamgoth: Initially made this update because the delegate components were a mix of named and default exports. I personally prefer this pattern because it makes imports very clear and can be done on one line. 

But I'm open to other patterns. The bigger thing to me is consistency throughout the app, so we should talk about it for sure.

components/delegations/DelegateCard.tsx
rafinskipg: Let's redo the layout in the future at some point 🚀

components/delegations/index.ts
rafinskipg: I like named exports because they are easier to search through the codebase, and it does not leave unnamed functions. 

About the `export * from x` i don't have a strong opinion but sometimes i had circular dependencies because of this. At the end we can find easily any file with the autocomplete. So i think i will vote for separate files with named exports but not having index.ts files that do a aggregation

PR #107, opened 2021-07-16
Fix/delegate state update create

b-pmcg:
### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/295/create-delegate-modal-hangs-after-signing-tx

-

### What does this PR do?

After a creating a vote delegate, the setter for vote delegate was not being called because it was inside an event listener listening for `accounts/CHANGE`, which is not emitted.

Adds a setter for Vote Delegate that is called after the create delegate transaction is mined.

-

### Steps for testing:

1. Connect a wallet that is not a vote delegate
2. Run through the create process
3. See the confirmation modal with the green check

OR

run the delegates/me.tsx test (not included in this PR)

### DIFF
--- a/pages/delegates/me.tsx
+++ b/pages/delegates/me.tsx
@@ -20,7 +20,11 @@ import TxDisplay from 'components/delegations/modals/TxDisplay';
 
 const CreateDelegate = (): JSX.Element => {
   const bpi = useBreakpointIndex();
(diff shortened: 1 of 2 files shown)


adamgoth (approved):

ethanbennett:
@adamgoth love to see Decker in my inbox! 😂

PR #110, opened 2021-07-20
show vote delegate balance instead of vote proxy balance if user has …

b-pmcg:
…both

### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/324/vote-proxy-balance-is-shown-on-executive-page-instead-of-vote-delegate-balance-if-user-has-both

### What does this PR do?

Problem:
A user reported that their vote proxy balance was being shown on the executive page even after they created a vote delegate contract.

Fix:
In the use `useLockedMkr` we now only show the vote proxy balance if the user does *not* have a vote delegate contract. Otherwise we show the vote delegate contract balance (ie maker locked in chief by vote delegate address).

Also updates the `useLockedMkr` hook in the other places it is used.

### Steps for testing:

1. Start with an account that has both a vote proxy & vote delegate contract
2. View the executive page and notice the balance shown is your vote delegate balance

The user who reached out to me offered to do further testing after we implemented a fix, I'll reach out to them since I can't get a vote proxy created on v1 anymore.

### Additional information:

**We should probably align with each other if this is the preferred behavior before merging**

### DIFF
--- a/components/executive/VoteModal.tsx
+++ b/components/executive/VoteModal.tsx
@@ -49,7 +49,7 @@ const VoteModal = ({ close, proposal, currentSlate = [] }: Props): JSX.Element =
 
   const addressLockedMKR =
     voteDelegate?.getVoteDelegateAddress() || voteProxy?.getProxyAddress() || account?.address;
-  const { data: lockedMkr } = useLockedMkr(addressLockedMKR, voteProxy);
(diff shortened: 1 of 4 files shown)


b-pmcg:
> This approach makes sense to me. At least until we implement some sort of delegate/proxy/account toggle feature in the future (still TBD).

Cool I confirmed it works with the reporting user

PR #114, opened 2021-07-24
WIP : Feat/UI toggle delegate

rafinskipg:
Things to do:

- Reflect on the UI of create delegate page that there is a behaviour change once you act as a delegate or not.
- Add a toggle for switching between delegate or normal in the Account modal and in the create delegate page. 
- Change the the account displayed on the account button when switching to a delegate. 
- Change the methods that call to the voting contracts when activating delegate

// Connect wallet button: Change the address once acting as delegate 
//   const [account, isActingAsDelegate] = useAccountsStore(state => [state.currentAccount, state.isActingAsDelegate]);
// Modal account : Add toggle delegate if has a delegate


rafinskipg:
I gotta fix the tests

PR #118, opened 2021-07-27
Feat/dark mode

b-pmcg:
### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/85/darkmode-for-vote-portal

### What does this PR do?

Adds a button to toggle dark mode and updates the theme with dark mode color tokens

### Steps for testing:

1. Use dark mode toggle and look around

Screen Shot 2021-07-28 at 4 22 39 PM

Screen Shot 2021-07-28 at 4 23 29 PM

Screen Recording 2021-07-28 at 1 09 33 PM


rafinskipg:
- When you load delegates / executive  /polling page, there is a loading placeholder that looks all white and the text is dark. 
- connect wallet state when you hover
- connect wallet modal when you open

We'll need to adjust some colors with time but as part of the design.  But otherwise looking great!

b-pmcg:
> * When you load delegates / executive  /polling page, there is a loading placeholder that looks all white and the text is dark.
> * connect wallet state when you hover
> * connect wallet modal when you open
> 
> We'll need to adjust some colors with time but as part of the design. But otherwise looking great!

All these should be fixed now

rafinskipg:
This can be fixed in the future: 
image


We need to add all the icons in the icon component so they can use the theme properties. It would be good to track it with a ticket

rafinskipg:
image

adamgoth (approved):
This looks 🔥 

The only small feedback I have is open modals are a bit hard to distinguish in dark mode.

b-pmcg:
> This can be fixed in the future:
> image
> 
> We need to add all the icons in the icon component so they can use the theme properties. It would be good to track it with a ticket

They actually are, but the portal does not use the icons from the package for some reason. Also the portal uses an older version of the icons package, I'm afraid to update because I don't know what will break, but we should make a ticket for that.

I really wish we could create tickets automatically from github comments, I wonder if... 🤔

in any case, this issue is fixed by adding `fill="currentColor" to the svg path"

PR #136, opened 2021-08-03
Feat/delegates system info

rafinskipg:
### Link to Clubhouse story
https://app.clubhouse.io/dux-makerdao/story/373/add-delegate-info-to-sidebar

### What does this PR do?

- Modifies API to add stats response
```
"stats":{"total":17,"shadow":16,"recognized":1,"totalMKRDelegated":9.5939}}
```
- Brings stats from the server side to the delegates system component
- Adds the delegates system info in the delegates page

### Steps for testing:
- Go to delegates page and see the component
- Load /api/delegates to get stats
-

image

### DIFF
--- a/__tests__/pages/delegates.tsx
+++ b/__tests__/pages/delegates.tsx
@@ -164,10 +164,10 @@ describe('Delegate Create page', () => {
     click(closeUndelegateBtn);
 
     // Voting weights are returned to 0 after undelegating
-    const newTotal = await screen.findByText(/Total MKR delegated/);
+    const newTotal = await screen.getByTestId('total-mkr-delegated');
     const newByYou = await screen.findByText(/MKR delegated by you/);
 
-    expect(newTotal.previousSibling).toHaveTextContent('0.00');
+    expect(newTotal).toHaveTextContent('0.00');
     expect(newByYou.previousSibling).toHaveTextContent('0.00');
   });
 });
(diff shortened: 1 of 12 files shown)


__tests__/pages/delegates.tsx
```diff
@@ -151,10 +164,10 @@ describe('Delegate Create page', () => {
     click(closeUndelegateBtn);
 
     // Voting weights are returned to 0 after undelegating
-    const newTotal = await screen.findByText(/Total MKR delegated/);
+    const newTotal = await screen.getByTestId('total-mkr-delegated');
```
adamgoth, line 167: I think it's better to find something by the text since that's what the user sees. Test ID might not change but it doesn't mean what the user is seeing hasn't changed in some future code change.

adamgoth (commented):
Looks good but I think 1) the spacing between sections should be consistent (delegate system info needs more margin on top) and 2) the MKR delegated should maybe have two decimals since that's what we show in the cards. 

What do you think?

__tests__/pages/delegates.tsx
rafinskipg: Actually testIDs are better than text , text is not only limited to one place, data-testids can be unique.  They are more "resilient" to change. A simple change on a copy may made your test unreliable.

rafinskipg:
> Looks good but I think 1) the spacing between sections should be consistent (delegate system info needs more margin on top) and 2) the MKR delegated should maybe have two decimals since that's what we show in the cards.
> 
> What do you think?

Yes, I will fix that and add some tests for the delegates system info box

__tests__/pages/delegates.tsx
rafinskipg: In tis case there is now 2 components with the same text, one for system info and one for the delegate

adamgoth (approved):

Issue #98, opened 2021-07-13
RFC: Code Module structure


rafinskipg:
## Code reorganization proposal.

The idea behind this proposal is to have a structure of code that is organized by responsibilities. 

/modules folder would replace /lib folder. Inside each module we would have:
- hooks: for client side hooks
- components: components related to this module
- api: fetch calls using web3 or other fetching mechanisms
- context: in case context are needed
- store: in case store is needed

Each module folder should have it's own tests contained.

```
/modules
  /esmodule
     /hooks
        /useESModuleData.tsx
     /components
        /ComponentSomething.tsx
        /__tests__
            /ComponentSomething.spec.tsx
     /types
        /esmodule.d.ts
     /api
        /fetchESModuleData.ts
        /__tests__
            /fetchESModuleData.spec.ts
     /esmodule.constants.ts
  /app
     /components
        /Button.tsx
     /hooks
       /useInitApp.tsx
  /account
      /hooks
         /useConnectAccount.tsx
      /lib
          /__tests__
             /mergeAccount.spec.ts
         /mergeAccounts.ts
  /maker
     /api
        /fetchMKRBalance.ts
        /fetchMKRStacked.ts
        /fetchMKRVoting.ts
     /hooks
         /useMKRBalance.tsx
         /useMKRLockeed.tsx
...
/pages
   /esmodule.tsx
   /...



```

adamgoth:
I like this pattern in general. It worked well in a large project I previously worked on. I think the challenge is breaking down the app into the right "modules" since we use various hooks, API calls, etc across the app.

b-pmcg:
I'm actually all for restructuring the code structure. The existing architecture was created long ago and maybe iterated over enough times where it no longer makes sense. Plus none of us had input on that.

I haven't had time to contemplate your specific suggestion here @rafinskipg but maybe its something we could all have a conversation about during a planning meeting after the MVP release.

b-pmcg:
I did a little more research on code structuring best practices for nextjs projects and found others recommending the modules approach. I like the idea behind it, one suggestion (from [here](https://dev.to/vadorequest/a-2021-guide-about-structuring-your-next-js-project-in-a-flexible-and-efficient-way-472)) would be to add a `common` directory for code that doesn't fit into a specific module, reducing a bit of the "thought complexity" surrounding module structure. That way we get to add code quickly when necessary, but retain the benefit of thoughtful organization of modules.

b-pmcg:
[...]

PR #146, opened 2021-08-11
Refactor/modules and add stats module

rafinskipg:
### What does this PR do?

- Adds the /api/stats/address endpoint that will return the votes for a certain address on an API level.
- Refactors old code into modules, (not all of it, is a progressive approach). This is just the /lib/api file that contained logic for different things, i splitted that into modules and in future PRs we can migrate more things from lib

See RFC #98 for module structure

image


components/header/VotingWeight.tsx
```diff
@@ -2,7 +2,7 @@ import { Flex, Text } from 'theme-ui';
 import useSWR from 'swr';
 import useAccountsStore from 'stores/accounts';
 import getMaker from 'lib/maker';
-import { getVotingWeightCopy } from 'lib/polling/getVotingWeightCopy';
+import { getVotingWeightCopy } from 'modules/polls/lib/getVotingWeightCopy';
```
adamgoth, line 5: I don't have very strong feelings about this but I feel like `utils` or `helpers` could be a better directory name for `lib` within these modules. The functions inside seem mostly to be utility/helper functions. Any thoughts?

modules/blog/api/fetchBlogPosts.ts
```diff
@@ -0,0 +1,29 @@
+import { BlogPost } from '../types/blogPost';
+
+
+export async function fetchBlogPosts(): Promise<BlogPost[]> {
+  // these are not just mocks anymore, but the actual blog posts that should be shown in production
+  return require('./mocks/blogPosts.json');
```
adamgoth, line 6: I know you didn't write this, just moved the code, but we should probably update this so our actual blog posts aren't coming from a mocks file 🤔

modules/stats/api/fetchStats.ts
```diff
@@ -0,0 +1,23 @@
...
+    },
+    votes: [
+      {
+        date: Date.now(),
+        address: '',
+        value: 'yes'
+      }
+    ]
+  });
+}
```
adamgoth, line 23: Maybe add a comment that this is a placeholder for now? The actual API hasn't been figured out yet.

adamgoth (approved):
I dig it in general. I left a few thoughts.

modules/blog/api/fetchBlogPosts.ts
rafinskipg: done

PR #151, opened 2021-08-19
Check for zero hash before fetching slate addresses

adamgoth:
### Link to Clubhouse story

https://app.clubhouse.io/dux-makerdao/story/251/metamask-vm-execution-error-repeated-in-console

### What does this PR do?

Reduces number of `MetaMask - RPC Error: invalid opcode: opcode 0xfe not defined` errors in the console.

### Steps for testing:

On this branch, go to the home page and executive page. If you are connected to an account that has not voted on proposals, you should not see the error. If you have voted for a proposal, you will see the error once every minute or so.

### Additional info:

The issue is caused by calling the `getSlateAddresses` in dai.js, which in turn calls the `slates` mapping in the Chief contract.

When a user is connected, we check the proposals they have voted on by getting their slate hash. The slate hash references a slate that contains an address of spells the user supports at the current time.

In the app's current state, we are making calls to `getSlateAddresses` even if the slate hash we received back was `0x0..`. This will always result in the `0xfe not defined` error. By putting in a check for the zero hash before calling `getSlateAddresses` with it, we can prevent these errors.

However, if the slate hash is not a zero hash, we then have to iteratively call `slates` (this is done in dai.js). We cannot tell how many addresses a slate may contain from the hash, so we iterate until it errors. That is why you will see an error returned once if a user has a value for their current vote slate.


b-pmcg:
interesting, that `getSlateAddresses` function is a weird one. So we just call it recursively until it errors?

b-pmcg (approved):

tyler17 (approved):

adamgoth:
> interesting, that `getSlateAddresses` function is a weird one. So we just call it recursively until it errors?

I guess it would be more accurate to say it recursively concatenates the next slate address until it gets an error. Then returns the concatenated array results up to that point. So we'll always get that last error in the console.

PR #155, opened 2021-08-24
Refactor/address api

rafinskipg:
### What does this PR do?

- Creates /api/address/[address] endpoint, it still returns delegate information if it's a delegate
- Creates route /address/[address] accessible for all delegates or normal addresses
- Creates page /address/[address] that shows delegate detail or address detail
- Adds "view profile" button to all delegates, so we can access the vote history of any delegate

### Steps for testing:
- Go to delegates and still can access the delegate detail
- Go to executive supporters and go to the address detail

image
image
image


next.config.js
```diff
@@ -26,5 +26,14 @@ module.exports = {
...
+  },
+
+  async rewrites() {
+    return [
+      {
+        source: '/delegates/:address',
+        destination: '/address/:address',
+      },
+    ];
+  },
```
adamgoth, line 38: Is this in case someone tries to access via an old URL?

pages/executive/[proposal-id].tsx
```diff
@@ -281,28 +270,29 @@ const ProposalView = ({ proposal }: Props): JSX.Element => {
...
-                            sx={{ color: 'accentBlue', fontSize: 3, ':hover': { color: 'blueLinkHover' } }}
-                          >
-                            {cutMiddle(supporter.address)}
-                          </Text>
-                        )}
-                      </ExternalLink>
+                        <ThemeUILink sx={{ mt: 'auto' }} title="Profile details">
+                          {supporter.name ? (
+                            <Text
+                              sx={{ color: 'accentBlue', fontSize: 3, ':hover': { color: 'blueLinkHover' } }}
```
Comment (adamgoth) [inline pages/executive/[proposal-id].tsx:283]: For some reason with this, on hover, the cursor is a text cursor and it doesn't show the URL down on the bottom of the page. I would expect the same behavior as the other links we use around the app (pointer cursor and a URL preview at the bottom of the window).

adamgoth (commented):
I think this API makes sense given the direction of the address history work we're doing. I think we should create a vote history feature branch and merge this into there for now, because in its current state, the address page is not complete. This would allow us to push develop branch to master without these changes if this work doesn't get completed by the end of the sprint. Thoughts?

next.config.js
rafinskipg: yeah, i leave it for now in case some delegate has their profile linked somewhere with the delegates address.

pages/executive/[proposal-id].tsx
Comment (rafinskipg) [reply pages/executive/[proposal-id].tsx:283]: Weird, i'll check it

Comment (rafinskipg) [reply pages/executive/[proposal-id].tsx:283]: ok fixede

PR #158, opened 2021-08-26
Style/add vote history UI

rafinskipg:
### What does this PR do?
- Adds tabs on the delegate detail
- Shows some information on the participation metrics
- Refactors poll vote history into polls module
- Refactors delegates into delegates module
- Adds little hover card https://www.loom.com/share/4e22fa7734fc4249943bb1aa744c5e39
- Another on the vote history detail: https://www.loom.com/share/b6870b327b264fe08abec56194dc22bf
-
### Steps for testing:
- Go to a delegate detail and start seeing the tabs

image
image
image

- Vote history is not finished
- Components for displaying results are not done


modules/delegates/components/DelegateDetail.tsx
```diff
@@ -20,9 +25,28 @@ export function DelegateDetail({ delegate, stats }: PropTypes): React.ReactEleme
 
   const { voteDelegateAddress } = delegate;
 
+
+  const tabTitles = [
+    delegate.status === DelegateStatusEnum.recognized ? 'Delegate credentials' : null,
+    'Participation metrics',
+    'Voting History'
+  ].filter(i => !!i) as string[];
```
adamgoth, line 33: We should be consistent with casing, so capitalize `Delegate Credentials` and `Participation Metrics`

modules/polls/components/PollVoteResultsCompact.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Text
+            variant="secondary"
+            color="onSecondary"
+            sx={{ textTransform: 'uppercase', fontSize: 1, fontWeight: 'semiBold', ml: 1, mr: 1}}
+            as="p"
+          >
+            No
+          </Text>
+        </Box>
+      </Box>
```
adamgoth, line 82: Feels off to have vote options hard-coded, we should be able to get these dynamically for each poll. Occasionally there are different options than these three which would make this component not compatible for those.

rafinskipg: Yep, this component is not finished, because the design really doesn't fit real requirements. 

Design is limited to this 3 options but polls can have many, then this component should be re-designed

adamgoth (commented):
I think this is from my changes on a previous branch commit and not this PR but noticed the address overflows on mobile:
Screen Shot 2021-08-30 at 11 15 18 AM

Also we should not show a border on the last result in vote history:
Screen Shot 2021-08-30 at 11 31 01 AM

modules/delegates/components/DelegateVoteHistory.tsx
```diff
@@ -0,0 +1,20 @@
+import { PollVoteHistoryList } from 'modules/polls/components/PollVoteHistoryList';
+import { AddressAPIStats } from 'modules/address/types/addressApiResponse';
+import { Box, Divider, Text } from 'theme-ui';
+import { Delegate } from '../types';
+
+export function DelegateVoteHistory({ delegate, stats }: { delegate: Delegate, stats: AddressAPIStats}): React.ReactElement {
+  return (
+    <Box p={[3,4]}>
+      <Text as="p" sx={{
+        fontSize: '18px',
```
adamgoth, line 10: In general, I think we should stick to our font-size variables from our theme spec.

modules/delegates/components/DelegateVoteHistory.tsx
```diff
@@ -0,0 +1,20 @@
+import { PollVoteHistoryList } from 'modules/polls/components/PollVoteHistoryList';
+import { AddressAPIStats } from 'modules/address/types/addressApiResponse';
+import { Box, Divider, Text } from 'theme-ui';
+import { Delegate } from '../types';
+
+export function DelegateVoteHistory({ delegate, stats }: { delegate: Delegate, stats: AddressAPIStats}): React.ReactElement {
```
adamgoth, line 6: Is `delegate` prop needed here?

modules/polls/components/PollVoteHistoryItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+            color="onSecondary"
+            sx={{ textTransform: 'uppercase', fontSize: 1, fontWeight: 'semiBold', textAlign: ['left', 'right'] }}
+            as="p"
+          >
+            Voted
+          </Text>
+          <Text as="p" sx={{ textAlign: ['left', 'right'] }}>
+            {vote.optionValue}
+          </Text>
+        </Box>
```
adamgoth, line 81: I think we still need to figure out how to show ranked choice options but based on the design, we should color code yes (green), no (red), and abstain (I think stays the default color).

modules/polls/components/PollVoteHistoryItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+import Tooltip from 'components/Tooltip';
+import { getNetwork } from 'lib/maker';
+import moment from 'moment';
+import Link from 'next/link';
+import { Box, Text, jsx, Link as ThemeUILink } from 'theme-ui';
+import { PollVoteHistory } from '../types/pollVoteHistory';
+import { PollVoteResultsCompact } from './PollVoteResultsCompact';
+import { parseRawPollTally } from '../helpers/parseRawTally';
+
+export function PollVoteHistoryItem({ vote }: { vote: PollVoteHistory }): React.ReactElement {
```
adamgoth, line 12: It looks like we're missing the Discussion link here from the design.

rafinskipg:
> I think this is from my changes on a previous branch commit and not this PR but noticed the address overflows on mobile:
> Screen Shot 2021-08-30 at 11 15 18 AM
> 
> Also we should not show a border on the last result in vote history:
> Screen Shot 2021-08-30 at 11 31 01 AM

The overflow happens in production too, can be fixed in further PRs.

The border on the last poll will be removed when i add the "Load more polls" functionality in a new PR

modules/delegates/components/DelegateVoteHistory.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <Box p={[3,4]}>
+      <Text as="p" sx={{
+        fontSize: '18px',
+        fontWeight: 500
```
adamgoth, line 11: `fontWeight: 500` can be `fontWeight: 'semiBold'`

modules/delegates/components/DelegateParticipationMetrics.tsx
```diff
@@ -0,0 +1,54 @@
...
+          {delegate.combinedParticipation ?? 'Untracked'}
+        </Text>
+      </Box>
+      <Box sx={styles.row}>
+        <Text as="p" sx={styles.text}>
+          Communication
+        </Text>
+        <Text as="p" sx={styles.text} ml={2}>
+          {delegate.communication ?? 'Untracked'}
+        </Text>
```
adamgoth, line 42: These look like they should be a different color per the design (a shade of grey instead of black like the title)

modules/delegates/components/DelegateParticipationMetrics.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </Box>
+      <Box sx={styles.row}>
+        <Text as="p" sx={styles.text}>
+          MKR Delegated
```
adamgoth, line 46: `MKR Delegated` isn't one of the participation metrics. I do think we should show this on the credentials page.

Design:
Screen Shot 2021-08-30 at 11 54 26 AM

modules/polls/components/PollVoteHistoryItem.tsx
rafinskipg: I know, is not finished. Will be added in new PRs

modules/delegates/components/DelegateVoteHistory.tsx
rafinskipg: @adamgoth  do you know how to get a 18px font-size with our theme spec?

modules/delegates/components/DelegateVoteHistory.tsx
rafinskipg: It will, because we need to add more components like the MKR delegated chart over time and so on. Probably those components can load asynchronously by calling the API with the delegate address

modules/polls/components/PollVoteHistoryItem.tsx
rafinskipg: yes, will be added

modules/delegates/components/DelegateVoteHistory.tsx
adamgoth: This has come up before. I think @b-pmcg pointed this out not too long ago, but if 18px isn't available as a size in our spec, then maybe the question should be why are we using it in the design? That means it's not appearing anywhere else in the app. Maybe we should check with Tiago on a different size to use. Or just use your best judgment between what's available in the spec.

PR #160, opened 2021-08-30
Vote poll address breakdown

adamgoth:
### What does this PR do?

Adds breakdown of poll results by address

### Steps for testing:

Go to poll detail page, click vote breakdown, see new address section

Screen Shot 2021-09-01 at 11 39 07 AM


lib/fscache/index.ts
```diff
@@ -18,6 +18,8 @@ export const fsCacheDel = (path: string): void => {
 };
 
 export const fsCacheGet = (name: string, expiryMs?: number): any => {
```
rafinskipg, line 20: 👍

lib/hooks/useDelegateAddressMap.ts
```diff
@@ -0,0 +1,32 @@
...
+  data: Record<string, string>;
+  loading: boolean;
+  error?: Error;
+};
+
+export const useDelegateAddressMap = (): DelegateAddressMapResponse => {
+  const { data, error } = useSWR(
+    '/delegate-address-map',
+    async () => {
+      const delegatesResponse = await fetchDelegates(getNetwork());
```
rafinskipg, line 15: I would change this call with a useSwr call over the API

```
const { data } = useSWR<DelegateAPIResponse>('/api/delegates?network=${getNetwork()}``)
```

rafinskipg: Also we can move this hook to delegates module

modules/polls/components/VoteWeightVisual.tsx
```diff
@@ -0,0 +1,113 @@
...
+  }, [ref.current]);
+
+  return (
+    <div ref={ref}>
+      <CirclesSvg diameter={width} poll={poll} tally={tally} />
+    </div>
+  );
+};
+
+export default VoteWeightVisual;
```
rafinskipg, line 113: 👍

modules/polls/types/pollTally.d.ts
```diff
@@ -13,22 +13,30 @@ type Result = {
...
+  eliminated: boolean;
+};
 
 export type RawPollTally = {
   winner: string | null;
   rounds: number;
   totalMkrParticipation: BigNumber;
   numVoters: number;
-  options: Record<number, PollTallyOption>
+  votesByAddress: PollTallyVote[];
```
rafinskipg, line 27: I'm thinking that we can introduce this as a new endpoint in a separate model.
 - /api/poll/[poll-id]/votes
Why?

Because we are using this model also in the vote poll history. Where we fetch like 50 polls. If we also have to fetch the voteByAddress for each poll is going to be a lot of info.  Also , if we split we can do async fetching of the vote stats. 

An alternative would be leaving it like this and making it an optional value in the model, so we don't have to fetch this when we are fetching a list of polls.

PR #186, opened 2021-09-22
Delegate modal alert

tyler17:
### Link to Clubhouse story
https://app.shortcut.com/dux-makerdao/story/371/alert-users-they-need-to-withdraw-mkr-from-chief-to-delegate-it
### What does this PR do?
Allows a user to easily withdraw MKR from chief before depositing into a delegate contract.
Also two bug fixes:
1. Show the correct MKR balance in the chief withdrawal modal for users with a delegate contract.
1. disable button when inputting a number larger than your MKR balance in delegate deposit and chief withdraw modals.

started adding a new yellow-colored alert, but instead just using the orange 'notice' one.
Screen Shot 2021-09-22 at 1 10 55 PM
Screen Shot 2021-09-22 at 1 11 00 PM
Screen Shot 2021-09-22 at 1 13 49 PM

### DIFF
--- a/lib/theme.js
+++ b/lib/theme.js
@@ -27,9 +27,6 @@ export default {
     ornament: '#000',
     skeletonColor: '#EEE',
     skeletonHighlightColor: '#F5F5F5',
-    info: '#fbd16f',
-    infoAlt: '#fff8eb',
-    onInfo: '#826417',
     modes: {
       dark: {
         primary: '#1DC1AE',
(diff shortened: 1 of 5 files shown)


lib/theme.js
```diff
@@ -145,6 +148,15 @@ export default {
...
   },
+  alerts: {
+    ...theme.alerts,
+    information: {
+      variant: 'alerts.primary',
+      borderColor: 'info',
+      bg: 'infoAlt',
+      color: 'onInfo'
+    }
+  },
```
adamgoth, line 159: I think info and notice are the same thing. Check the "Badges and Alerts" section here: https://dai-ui.vercel.app/

adamgoth (approved):
LGTM. The modals on top of modals is kind of a weird experience but it works for this flow.

I think you can remove the new alert you started adding, I think the existing "notice" one is the right one to use.

PR #199, opened 2021-10-08
Add call to new plurality function

b-pmcg:
### What does this PR do?

Fixes issue where UI didn't display a winner (example of issue: https://vote.makerdao.com/polling/QmNzDxBK?network=mainnet#poll-detail)

### Steps for testing:

Go to this poll and see that the correct winner is displayed (`/polling/QmNzDxBK?network=mainnet#poll-detail`)

Go to other random polls, make sure everything looks good in results and breakdown

Issue was caused by running all results through the IRV algorithm. Fix includes only running ranked choice polls through this algorithm, otherwise sorting by most MKR support to find winners for plurality votes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@makerdao/dai": "^0.32.14-beta0",
-    "@makerdao/dai-plugin-governance": "^0.14.2-beta0",
+    "@makerdao/dai-plugin-governance": "^0.14.2",
     "@makerdao/dai-plugin-ledger-web": "^0.9.10",
     "@makerdao/dai-plugin-mcd": "^1.6.36-beta0",
     "@makerdao/dai-plugin-trezor-web": "^0.9.7",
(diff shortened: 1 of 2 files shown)


modules/polling/helpers/parseRawTally.ts
```diff
@@ -36,10 +37,31 @@ export function parseRawPollTally(rawTally: RawPollTally, poll: Poll): PollTally
...
+        winner: rawTally.options?.[key]?.winner ?? false
+      } as PluralityResult;
+    })
+    .sort((a, b) => {
+      const valueA = a.mkrSupport;
+      const valueB = b.mkrSupport;
+      if (valueA.eq(valueB)) return a.optionName > b.optionName ? 1 : -1;
+      return valueA.gt(valueB) ? -1 : 1;
+    });
+
```
b-pmcg, line 60: This is a fairly ugly way to handle this, but I couldn't figure out how to do it in a single `.map` and make typescript happy.

PR #215, opened 2021-11-01
Add fetch for categories json on validation

adamgoth:
### Link to Clubhouse story

https://app.shortcut.com/dux-makerdao/story/481/update-categories-in-the-validator-function-for-when-polls-are-created

### What does this PR do?

When validating poll markdown (`polling/create`), this update will now fetch categories dynamically from the community repo. If it fails, it will fallback to the hardcoded categories we have been relying on.

### Steps for testing:

Go to `polling/create`, ensure validation works

### DIFF
--- a/modules/polling/helpers/validator.ts
+++ b/modules/polling/helpers/validator.ts
@@ -7,7 +7,7 @@ import { POLL_VOTE_TYPES_ARRAY } from '../polling.constants';
 
 // find the most up-to-date list here:
 // https://github.com/makerdao/community/blob/master/governance/polls/meta/categories.json
-const validCategories = [
+const hardcodedCategories = [
   'Collateral',
   'Oracles',
   'Governance',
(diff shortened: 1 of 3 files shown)


modules/polling/helpers/validator.ts
```diff
@@ -32,18 +32,32 @@ type ValidationResult = {
...
+}
+
 export async function validateUrl(url: string, poll?: PartialPoll): Promise<ValidationResult> {
   const resp = await fetch(url);
   const text = await resp.text();
-  const result = validateText(text);
+
+  const liveCategories = await fetchCategories();
+  // fallback to hardcoded categories if live category fetch fails
+  const result = validateText(text, liveCategories ?? hardcodedCategories);
```
rafinskipg, line 52: maybe return hardcodedCategories from "fetchCategories" if the requests fails

rafinskipg (changes requested):

modules/polling/helpers/validator.ts
rafinskipg: Note: this is not a "hard" request changes comment. the PR LGTM

rafinskipg (approved):

PR #233, opened 2021-11-05
add colormode script sha back to _app.tsx

rafinskipg:


b-pmcg:
@rafinskipg we don't need this PR any more, correct?

PR #235, opened 2021-11-08
Fix/tooltip delegates

rafinskipg:
Adds a little tooltip on participation metrics for delegates


pages/executive/[proposal-id].tsx
```diff
@@ -254,7 +254,7 @@ const ProposalView = ({ proposal }: Props): JSX.Element => {
                 tabListStyles={{ pl: [3, 4] }}
                 tabTitles={['Spell Details']}
                 tabPanels={[
-                  <div key={1} sx={{ p: [3, 4] }}>
+                  <div key={'spell'} sx={{ p: [3, 4] }}>
```
Comment (adamgoth) [inline pages/executive/[proposal-id].tsx:257]: Should these `key` changes be included?

modules/delegates/components/DelegateCard.tsx
```diff
@@ -33,22 +37,6 @@ export function DelegateCard({ delegate }: PropTypes): React.ReactElement {
...
-  );
-  const communicationTooltipLabel = (
-    <>
-      The percentage of votes for which the delegate has publicly <br />
-      communicated their reasoning in addition to voting. <br />
-      Combines stats for polls and executives. <br />
-      Updated weekly by the GovAlpha Core Unit. <br />
-    </>
-  );
-
```
adamgoth, line 51: I moved these to the participation metrics component so we can reuse label text

adamgoth (approved):
LGTM but wondering if these `key` changes should be included?

pages/executive/[proposal-id].tsx
Comment (rafinskipg) [reply pages/executive/[proposal-id].tsx:257]: they don't hurt, i almost prefer them to the indexed key that can fail when moving components

Comment (adamgoth) [reply pages/executive/[proposal-id].tsx:257]: Just wanted to make sure if it was intended for this PR but I agree 👍

PR #241, opened 2021-11-09
Resolve ens addresses

rafinskipg:
### Link to Clubhouse story
https://app.shortcut.com/dux-makerdao/story/697/start-using-ens-resolvers-throughout-the-gov-portal

### What does this PR do?
- Allows to navigate to addresses using ens: 
Examples: `/delegates/mkrdelegate.schuppi.eth` or `/address/robocopsgonemad.eth`

- In the table of addresses voting for a poll, renders address as ens if resolve

- In the table of executive supporters shows the ens names

### Steps for testing:
Navigate through the app 

image
image

- There are some caveats to this story. I would like that ENS were resolved on any data structure coming from the backend, for example: 
```
{
  address: '0x0123213',
  displayName: 'address.eth'
}
```

This data structure is generic enough that we can use for delegates and any address. I'll try to work on that after this


modules/polling/components/VotesByAddress.tsx
```diff
@@ -48,13 +49,15 @@ const VotesByAddress = ({ tally, poll }: Props): JSX.Element => {
           {votes ? (
             <>
               {votes.map((v, i) => (
-                <tr key={i}>
+                <tr key={i} style={{ fontSize: bpi < 1 ? '12px' : '16px' }}>
```
adamgoth, line 52: We should stick to our theme UI variables for font sizes

adamgoth (approved):
Looks good!

One thing I noticed, not sure what's causing it so not sure if it's related to this PR for sure, but every time you go to `/address/<123>` there's a console warning like this:

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval' 'sha256-a0L6Pfwt+Nftvey0NflqMTGt/tO5UMFmI/PAiNnoYRo='". Either the 'unsafe-inline' keyword, a hash ('sha256-ffXIVJIqwbhUsKZuw2L+l9gEWeLEqzkH/jgc2LNrgRA='), or a nonce ('nonce-...') is required to enable inline execution.```

File: modules/web3/ens.ts [hunk 2]
```diff
@@ -11,3 +11,13 @@ export async function getENS(address: string): Promise<string> {
...
 }
+
+export async function resolveENS(ensName: string): Promise<string> {
+  const maker = await getMaker();
+
+  const provider = maker.service('web3')._web3.currentProvider;
+
+  const ens = new ENS(provider);
+
+  return await ens.resolver(ensName).addr();
```
b-pmcg, line 22: FYI ethers has an ENS resolver we could use: https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods

it seems that the `ethereum-ens` package hasn't been updated in a while

b-pmcg (approved):
LFGTM!

tyler17 (approved):

rafinskipg:
> causing it so not sure i

yeah i think this happens in prod too

modules/polling/components/VotesByAddress.tsx
rafinskipg: how can i do 12px or 16px ? @tyler17  @b-pmcg @adamgoth ?

adamgoth: Reference the `fontSize` array here and use the index: https://github.com/makerdao/dai-ui/blob/master/packages/dai-ui-theme-maker-neue/lib/index.js

12px would be `fontSize: 1` and 16px would be `fontSize: 3` since those are their indexes in that array.

rafinskipg: thanks @adamgoth  ,

PR #243, opened 2021-11-10
add hw address pagination

tyler17:
### Link to Clubhouse story
https://app.shortcut.com/dux-makerdao/story/550/trezor-wallet-not-working-on-latest-deployment
### What does this PR do?
Loads 25 addresses instead of 10 for hardware wallets, and shows them over 5 pages of 5 addresses, rather than one large modal with 10 addresses
### Steps for testing:
Connect a hardware wallet, try selecting an address not on the first page.
Screen Shot 2021-11-09 at 10 19 55 PM

### Add a GIF:
dog

### DIFF
--- a/modules/app/components/layout/header/AccountSelect.tsx
+++ b/modules/app/components/layout/header/AccountSelect.tsx
@@ -56,6 +56,9 @@ const WrappedAccountSelect = (): JSX.Element => (
   </Web3ReactProvider>
 );
 
+const ADDRESSES_PER_PAGE = 5;
+const MAX_PAGES = 5;
+
 const AccountSelect = (): React.ReactElement => {
   const { setUserData } = useContext(AnalyticsContext);
 
@@ -95,6 +98,8 @@ const AccountSelect = (): React.ReactElement => {
   const [showHwAddressSelector, setShowHwAddressSelector] = useState(false);
   const [hwSelectCallback, setHwSelectCallback] = useState<(err: Error | null, address?: string) => void>();
 
+  const [hwPageNum, setHwPageNum] = useState(0);
+
   const close = () => setShowDialog(false);
   const bpi = useBreakpointIndex();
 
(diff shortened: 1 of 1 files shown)


rafinskipg (commented):
Looking good! 

Could we make the MAX_PAGES be relative to the number of addresses instead of fixed?  For example, do not show the buttons if there are no more than 5 addresses, and disable the buttons once you reach the last address in your wallet

tyler17:
Well, it's not like a HW wallet has a fixed number of addresses.  You can use any address you want, I think Trezor can handle on the order of billions addresses

rafinskipg:
> Well, it's not like a HW wallet has a fixed number of addresses. You can use any address you want, I think Trezor can handle on the order of billions addresses

But you already have the "addresses" item , so you know the length of the addresses array, right? Is this array containing all the addresses on that hardware wallet? 

If so, the pagination can be improved using that length, not a fixed number of pages. 

Am i missing something maybe?

tyler17:
The length of the `addresses` array is whatever number we pass in as the accountsLength on line 140, which I just hardcoded as 25. I chose 25 and not a larger number because it takes a while for ledger (ledger live specifically) to generate the addresses.  I think there's two improvements that can be made here:

1. Right now this only supports the ledger live path, not the legacy ledger path.  So someone who setup their ledger before ledger live launched (I think summer of 2018) won't be able to access their addresses.
2. For ledger live, it's still takes a few seconds to generate the 25 addresses, so for ledger live we could only grab 5 addresses at a time, and then only generate more if they go to the next page. This would also allow someone to access an address past the 25th one.

But I think we can address that in a different PR, I can create clubhouse tickets.

adamgoth (commented):
Code looks good to me but when I try to connect trezor, I get stuck on the loading screen. Have you seen this issue before?

Screen Shot 2021-11-15 at 11 10 31 AM

rafinskipg:
I connected with the trezor succesfully.

rafinskipg (approved):

PR #253, opened 2021-11-25
Fix polling date filter issue

rafinskipg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/551/poll-date-filter-filter-on-end-date-instead-of-creation-date

### What does this PR do?
Changes text and filters in poll page 

### Steps for testing:
Try filters on poll page, see if they make sense
image

### DIFF
--- a/modules/executive/components/DateFilter.tsx
+++ b/modules/executive/components/DateFilter.tsx
@@ -43,14 +43,14 @@ export default function DateFilter(props): JSX.Element {
     <FilterButton
       name={() => {
         if (!startDateDisplay && !endDateDisplay) return 'Date Posted';
-        if (!startDateDisplay) return `Date Posted: before ${endDateDisplay}`;
-        if (!endDateDisplay) return `Date Posted: after ${startDateDisplay}`;
-        return `Date Posted: ${startDateDisplay} - ${endDateDisplay}`;
+        if (!startDateDisplay) return `Date Filter: before ${endDateDisplay}`;
+        if (!endDateDisplay) return `Date Filter: after ${startDateDisplay}`;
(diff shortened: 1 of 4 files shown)


modules/executive/components/DateFilter.tsx
```diff
@@ -50,7 +50,7 @@ export default function DateFilter(props): JSX.Element {
       {...props}
     >
       <Grid gap={2} columns="max-content max-content" sx={{ alignItems: 'baseline' }}>
-        <Text>After</Text>
```
tyler17, line 53: Slightly tangential, up on line 45, should we change `Date Posted` to `Date Filter` to be consistent with the polling page?

tyler17 (approved):

modules/executive/components/DateFilter.tsx
```diff
@@ -59,7 +59,7 @@ export default function DateFilter(props): JSX.Element {
           />
         </Flex>
 
-        <Text>Before</Text>
+        <Text>Executive posted before:</Text>
```
tyler17, line 62: don't have a strong opinion, but feel this could be shortened to just `Posted Before` before and `Posted after`

modules/polling/components/DateFilter.tsx
```diff
@@ -45,7 +45,7 @@ export default function DateFilter(props): JSX.Element {
       {...props}
     >
       <Grid gap={2} columns="max-content max-content" sx={{ alignItems: 'baseline' }}>
-        <Text>After</Text>
+        <Text>Poll expired after:</Text>
```
tyler17, line 48: Same here, feel like it could be shortened to just `Ended after`, (I think end is more consistent with the wording elsewhere than expired)

PR #260, opened 2021-11-29
Test/add test empty tally

rafinskipg:
Adds a test to check that even an empty tally from dai.js do not break the data returned to the FE

### DIFF
--- a/modules/polling/api/fetchVotesByAddress.ts
+++ b/modules/polling/api/fetchVotesByAddress.ts
@@ -1,0 +1,17 @@
+import BigNumber from 'bignumber.js';
+import { SupportedNetworks } from 'lib/constants';
+import getMaker, { getNetwork } from 'lib/maker';
+import { PollTallyVote } from '../types';
+
+export async function fetchVotesByAddresForPoll(
+  pollId: number,
+  network?: SupportedNetworks
+): Promise<PollTallyVote[]> {
+  const currentNetwork = network || getNetwork();
+
+  const maker = await getMaker(currentNetwork);
+
+  return (await maker.service('govPolling').getMkrAmtVotedByAddress(pollId)).sort((a, b) =>
+    new BigNumber(a.mkrSupport).lt(new BigNumber(b.mkrSupport)) ? 1 : -1
+  );
+}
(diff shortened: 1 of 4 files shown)


modules/polling/api/fetchVotesByAddress.ts
```diff
@@ -0,0 +1,17 @@
...
+  network?: SupportedNetworks
+): Promise<PollTallyVote[]> {
+  const currentNetwork = network || getNetwork();
+
+  const maker = await getMaker(currentNetwork);
+
+  return (await maker.service('govPolling').getMkrAmtVotedByAddress(pollId)).sort((a, b) =>
+    new BigNumber(a.mkrSupport).lt(new BigNumber(b.mkrSupport)) ? 1 : -1
+  );
+}
```
rafinskipg, line 17: Separated into a file for easy mocking

modules/polling/helpers/__tests__/getPollTally.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('getPollTally', () => {
+  // 29-11-2021, the gov portal crashed due to a empty options object being returned to the UI and not handled
+  // correctly. We add this test to check that we always return a non-empty options object to the FE
+  it('Returns a correct tally even when the options are empty', async () => {
+    (fetchPollTally as jest.Mock).mockReturnValue(
+      Promise.resolve({
+        totalMkrParticipation: new BigNumber(0),
+        winner: '0',
+        numVoters: 0,
+        options: {}
```
rafinskipg, line 35: This simulates the dai.js empty options

b-pmcg: nice, love it. JW if we want to mention the crash in the comments?

b-pmcg (approved):

b-pmcg:
Good to have this test. It makes me wonder if we also should make the change in this PR or not, since it is handled in the FE: https://github.com/makerdao/dai.js/pull/313

modules/polling/helpers/__tests__/getPollTally.spec.ts
rafinskipg: i just want to leave some impression about why we have this kind of test, it can happen in the future other people look at this code and  don't think about this. But yeah I'll change it

rafinskipg:
> Good to have this test. It makes me wonder if we also should make the change in this PR or not, since it is handled in the FE now: [makerdao/dai.js#313](https://github.com/makerdao/dai.js/pull/313)

Maybe we don't need it.  :(

PR #261, opened 2021-11-29
Feat: Allow to send a query parameter to determine the community repo…

rafinskipg:
Allows to send a query parameter to the governance portal to fetch Polls Metadata from a different branch than master (on the governance community repo). 

This will allow to test new polls or metadata without having to merge into the community repo master branch ( which affects production)

To see a different branch use the following url, replacing "another" for any branch

`/polling/all-polls?branch=another`

## how to test it?

Go to `/polling/QmZmmLNv?network=goerli&branch=dependabot/npm_and_yarn/r2d/theme-ui-0.12.1#poll-detail` 

This is a branch from the community repo, in the url

https://github.com/makerdao/community/branches

### DIFF
--- a/modules/polling/api/fetchPolls.ts
+++ b/modules/polling/api/fetchPolls.ts
@@ -3,7 +3,7 @@ import { fsCacheGet, fsCacheSet } from 'lib/fscache';
 import getMaker, { isTestnet } from 'lib/maker';
 import { markdownToHtml } from 'lib/utils';
 import invariant from 'tiny-invariant';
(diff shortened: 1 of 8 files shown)


rafinskipg:
As @b-pmcg  commented, it does not make sense since the polls are activated at a certain time so they appear in production at a certain time,

b-pmcg:
@rafinskipg I've been trying to think of a way to leverage this idea, because the mechanics of it are good. Maybe we could utilize feature flags to remove the date check that prevents displaying polls before their activated time. Then LFW or someone could quickly look at the polls even though they are not "active" at the moment.

PR #262, opened 2021-12-01
fetch delegator data in API

b-pmcg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/625/add-list-of-delegators-to-delegate-profile-page

### What does this PR do?
Shows a table of delegators who have delegated to a delegate.

### Steps for testing:
1. go to a delegate page and see the table
2. find a delegate with 0 MKR and don't see the table

Screen Shot 2021-11-30 at 6 16 34 PM

### DIFF
--- a/lib/theme.js
+++ b/lib/theme.js
@@ -60,6 +60,8 @@ export default {
     tagColorFifteenBg: '#FFFBEF',
     tagColorSixteen: '#FF8237',
     tagColorSixteenBg: '#FFFBEF',
+    bull: '#1AAB9B',
+    bear: '#F77249',
     modes: {
       dark: {
         primary: '#1DC1AE',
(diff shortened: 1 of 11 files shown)


rafinskipg (approved):
Code looks great! 

but i think the UI needs a couple of small changes:
- the + button is smaller in the design
- the text should be in a higher font weight I think

In the design it says "voting weight" and in the PR "total percentage"

We need the links to verify the individual transactions ( that can be another PR)

b-pmcg:
> Code looks great!
> 
> but i think the UI needs a couple of small changes:
> 
> * the + button is smaller in the design
> * the text should be in a higher font weight I think
> 
> In the design it says "voting weight" and in the PR "total percentage"
> 
> We need the links to verify the individual transactions ( that can be another PR)

cool fixed the sizing issues, only thing is usually we use the term voting weight to refer to maker amount, I don't think its the right label for the percentage column. Maybe just "percentage" is better?

PR #267, opened 2021-12-08
account for multiple txs on same day

b-pmcg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/876/mkr-delegate-weight-chart-dates-on-chart-don-t-always-line-up-with-dates-on-the-delegation-history-table

### What does this PR do?
Handle situation where there are multiple transactions in a single day, selecting the most recent one which will have the correct lockTotal.

### Steps for testing:
Open the metrics tab and compare the delegation history table with the MKR weight chart and verify that the dates on the chart and the table line up.

**Before:**

**After:**
Screen Shot 2021-12-09 at 10 47 57 AM

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -5,6 +5,7 @@ import { cloneElement } from 'react';
 import { jsx } from 'theme-ui';
 import { css, ThemeUIStyleObject } from '@theme-ui/css';
 import BigNumber from 'bignumber.js';
+import { format } from 'date-fns';
 import { CurrencyObject } from 'modules/app/types/currency';
 import { SpellStateDiff } from 'modules/app/types/spellStateDiff';
 import { SupportedNetworks, ETHERSCAN_PREFIXES } from './constants';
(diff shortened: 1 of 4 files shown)


lib/utils.ts
```diff
@@ -176,3 +177,19 @@ export const sortBytesArray = _array =>
...
+    return isoDate;
+  }
+};
+
+export const formatIsoDateConversion = date =>
+  parseInt(
+    format(new Date(isoDateConversion(date)), 'D', {
+      useAdditionalDayOfYearTokens: true
+    })
+  );
```
adamgoth, line 195: We have a `lib/datetime` file that these may make more sense to be located in.

adamgoth (approved):
LGTM, left a suggestion about which util file to add the helpers to

rafinskipg (approved):
LGTM but please refer to @adamgoth  comment about the library for dates

PR #274, opened 2021-12-14
delete tally.options and use tally.results

b-pmcg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/828/remove-redundant-object-array-from-tally-calculation-results

### What does this PR do?
deletes the tally.options property so it can't be used anywhere in the code, since the "results" property has the same info and more.

### Steps for testing:
Look at poll vote breakdown and "winning option" box on poll cards for example.


pages/polling/[poll-hash].tsx
```diff
@@ -282,14 +282,13 @@ const PollView = ({ poll }: { poll: Poll }) => {
...
-                tally &&
-                tally.totalMkrParticipation > 0 && (
+                tally && tally.totalMkrParticipation > 0 ? (
                   <Box>
                     <Divider my={0} />
                     <PollWinningOptionBox tally={tally} poll={poll} />
                     <Divider my={0} />
                   </Box>
-                )
+                ) : null
```
Comment (b-pmcg) [inline pages/polling/[poll-hash].tsx:291]: I was getting a typescript compiler error because the previous implementation was potentially resolving to `false`, so I refactored it into a ternary and added `null` as a possible type in the Tabs.tsx component props. There were other ways to fix the type error, but this seemed cleanest to me. Open to suggestions though.

adamgoth (approved):
LGTM 💪

rafinskipg (approved):

PR #275, opened 2021-12-14
#809 Refresh votes when changing account

adamgoth:
### Link to Shortcut ticket:

https://app.shortcut.com/dux-makerdao/story/809/refresh-votes-when-changing-account

### What does this PR do?

Updates vote selections on polling page when an account changes 

### Steps for testing:

1. Go to polling page
2. Notice state of the votes
3. Swtich to an account with different vote selections
4. See updated votes ✨


rafinskipg (changes requested):
Is not working well with ranked choice options:

image

It should go back to the select component

adamgoth:
> Is not working well with ranked choice options:
> 
> image
> 
> It should go back to the select component

Good catch. Fixed this as well as an issue where the default value for single select wasn't reflected after change.

rafinskipg (approved):

PR #286, opened 2022-01-05
Chore/replace fire event with user event

adamgoth:
### What does this PR do?

Replaces `fireEvent` in tests with `userEvent`

This is recommended by react testing library: https://testing-library.com/docs/dom-testing-library/api-events

Why?

> the userEvent library uses fireEvent methods to trigger some events, but it calls them in sequence and in the order in which they would occur if a real user was the one behind the screen.

Article with more details: https://blog.mimacom.com/react-testing-library-fireevent-vs-userevent/

### DIFF
--- a/__tests__/helpers.tsx
+++ b/__tests__/helpers.tsx
@@ -1,4 +1,4 @@
-import { act, render, RenderResult, screen } from '@testing-library/react';
+import { render, RenderResult, screen } from '@testing-library/react';
 import { TestAccountProvider } from '@makerdao/test-helpers';
 import { formatAddress } from 'lib/utils';
 import { ThemeProvider } from 'theme-ui';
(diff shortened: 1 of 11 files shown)


rafinskipg (changes requested):
I was reading this issue about wrapping userEvent in "act()" and they say it should never be wrapped.

Please , take a look to https://github.com/testing-library/user-event/issues/497

adamgoth:
> I was reading this issue about wrapping userEvent in "act()" and they say it should never be wrapped.
> 
> Please , take a look to [testing-library/user-event#497 (comment)](https://github.com/testing-library/user-event/issues/497#issuecomment-729904582)

Good catch. Not necessary. Fixed and removed.

rafinskipg (approved):

PR #296, opened 2022-01-13
Hotfix/2022 jan 13

adamgoth:
### What does this PR do?

Fix issue with MKR weight over time chart for delegate

Bundle comment posts to db into a single tx

Fix ENS resolution issue


b-pmcg (commented):
still confused why the checks for install and test says "Expected — Waiting for status to be reported" here. I re-ran the pipeline from circleci and everything passed, so it appears to be an issue with github connecting to circleci maybe?

PR #297, opened 2022-01-13
Fix ens issue, use ethers instead

adamgoth:
Fixes 0x000 issue: Screen Shot 2022-01-13 at 7 05 26 AM

We can use ethers instead of an additional lib to fetch ENS data


rafinskipg (changes requested):
adam LGTM but i would leave the try / catch, what happens if the ENS lookup fails to resolve?

PR #298, opened 2022-01-13
Refactor dai.js out of system stats (WIP)

adamgoth:


modules/web3/hooks/useTotalDai.tsx
```diff
@@ -0,0 +1,37 @@
+import useSWR from 'swr';
+import { ethers } from 'ethers';
+import { useContracts } from 'modules/web3/hooks/useContracts';
+import { WAD, RAY } from 'modules/web3/web3.constants';
+
+export const useTotalDai = (): any => {
+  const { pot } = useContracts();
```
b-pmcg, line 7: can we have a hook called `usePot`? :P

lib/string.ts
```diff
@@ -6,3 +9,24 @@ export function cutMiddle(text = '', left = 6, right = 4): string {
   if (text.length <= left + right) return text;
   return `${text.substring(0, left)}...${text.substring(text.length - right, text.length)}`;
 }
+
+const builtInUnits = ['wei', 'kwei', 'mwei', 'gwei', 'szabo', 'finney', 'ether'];
+
+export function formatValue(value: BigNumber, type: string | number, dp = 2, withCommas = true): string {
```
b-pmcg, line 15: I added some comments about this function [here](https://app.shortcut.com/dux-makerdao/story/1025/create-helpers-for-parseether-parseunit-formatether-formatunit?vc_group_by=day&cf_workflow=500000005&ct_workflow=all)

modules/web3/hooks/useTotalDai.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    data,
+    loading: !error && !data,
+    error
+  };
+};
+
+// temp comment: dai.js implementation for reference
+
+// async getTotalDai() {
```
b-pmcg, line 20: I believe this is the total Dai in the DSR, `vat.debt()` is the total dai in circulation (I compared the result with system stats on prod and in daistats, and it's the same).

PR #300, opened 2022-01-14
Hooks for system stats (DSR needs help)

adamgoth:
All system stats are using hooks now, but the DSR one isn't working

App keeps crashing for me whenever I try to do `.pow(SECONDS_PER_YEAR)`


b-pmcg:
@adamgoth I spent wayyy too long trying to figure this out, but not getting anywhere. I looked at [daistats](https://github.com/nanexcool/daistats/blob/80cd2c47b718bae72155943c720ed2f5d47486c8/src/App.jsx#L1030) implementation and tried to reproduce that exactly, and it didn't work. I also reproduced the way we do it in dai.js (using BigNumber.js), which should be identical to how it currently works on prod.

My feeling is it has something to do with where we're doing it in the code, like maybe it should be in a separate util? Or maybe the hook is being called too many times & the POW operation gets too many threads going? I'm going to switch gears and we can come back to it later because I'm stumped.

b-pmcg (approved):
lgtm

PR #305, opened 2022-01-18
Add gql (poc)

adamgoth:
Uses `@graphql-codegen` to generate types based on our graphQL API schema. Fetches this dynamically from a provided endpoint URL.

Uses `graphql-request` to send graphql requests to the API.

Replaces dai.js for gql requests direct to server for:
* ESM staking history
* Fetching list of polls
* Fetching votes by address


rafinskipg (approved):

b-pmcg:
LGTM

tyler17:
Nice! So, to automatically update the schema and types do we just run `yarn gql-generate` (assuming whatever url in the codegen file has the updated schema)? @adamgoth

adamgoth:
> Nice! So, to automatically update the schema and types do we just run `yarn gql-generate` (assuming whatever url in the codegen file has the updated schema)? @adamgoth

Yep exactly. We'll have to remember to update the URL to the point to the updated schema when we want to update it (whether that's localhost or the staging URL probably)

PR #308, opened 2022-01-20
Networking updates

adamgoth:
Remove function that syncs maker object with account

Remove network from query params (just rely on active provider connection instead)

Add default provider wrapper when there is no connector injected so that `useActiveWeb3React` still works

Add connect chain name to header


modules/app/components/layout/header/NetworkAlertModal.tsx
```diff
@@ -73,6 +81,8 @@ const NetworkAlertModal = ({
               Your wallet is connected to an unsupported network, please switch it to{' '}
               {SupportedNetworks.MAINNET} or {SupportedNetworks.GOERLI} to continue.
             </Text>
+            {/* <Button onClick={() => handleSwitchNetwork(SupportedNetworks.MAINNET)}>Switch to mainnet</Button>
```
b-pmcg, line 84: nice, kinda happy to get rid of this. Also, just flagging the `network mismatch` error above, probably need to adjust the code there too.

b-pmcg (approved):
looks good, would it make sense to call the `chainIdToNetworkName` function inside the `useActiveWeb3React` hook and return the network name there?

adamgoth:
> looks good, would it make sense to call the `chainIdToNetworkName` function inside the `useActiveWeb3React` hook and return the network name there?

Had the same thought. I think it would make it cleaner to use for sure. I only hesitated since right now we just return the context unaltered from web3-react. But I don’t think it’s an issue to add in network too and return it.

b-pmcg:
also I love the network indicator, we should just leave it like that and not tell @tiago-varandas  xD

PR #309, opened 2022-01-21
Remove accounts store

rafinskipg:
# Remove store

- Remove zustand account store
- have useAccount() hook that returns the vote delegate, vote proxy, old vote proxy 

WIP: try to get the hot / cold vote proxy


modules/app/hooks/useAccount.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function useAccount(): UseAccountResponse {
+  const { account } = useActiveWeb3React();
+
+  if (!account) {
+    return {
+      account,
+      mutate: () => null
+    };
+  }
```
b-pmcg, line 37: So I think there is a problem with this implementation, we are conditionally rendering the hooks based on `account` which breaks one of the rules of hooks. If we don't have an account, we don't run the hooks below, but when we finally get it, we do. React doesn't like that.

I think we need to make the hooks handle the case of `!account` so we can run each of the hooks in the same order every time. One way to do this is with useSWR which does allow for [conditional fetching](https://swr.vercel.app/docs/conditional-fetching), eg: `const { data, error } = useSWR(contractAddress ? '${contractAddress}/vote-delegate-contract' : null, ...` or put the actual contract calls in a non-hook and return that conditionally, eg:`getChiefDeposits` here: 
```
const { data, error, mutate } = useSWR(
    addressToCache ? '${chief.address}/user/mkr-locked${addressToCache}' : null,
    async () => {
      return addressToCache ? await getChiefDeposits(addressToCache, chief) : null;
    }
  );
  ```

File: modules/app/hooks/useVoteProxyOldAddress.ts [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+  const { chainId } = useActiveWeb3React();
+
+  // Only works in mainnet
+  const network = chainIdToNetworkName(chainId as SupportedChainId);
+
+  if (
+    network === SupportedNetworks.GOERLI ||
+    network === SupportedNetworks.GOERLIFORK ||
+    network === SupportedNetworks.TESTNET
+  ) {
```
adamgoth, line 27: Maybe `!== SupportedNetworks.MAINNET` instead and put it earlier so it returns right away?

modules/executive/components/ExecutiveOverviewCard.tsx
```diff
@@ -29,14 +29,12 @@ type Props = {
...
   const { trackButtonClick } = useAnalytics(ANALYTICS_PAGES.EXECUTIVE);
-  const account = useAccountsStore(state => state.currentAccount);
+  const { account } = useAccount();
   const { data: mkrOnHat } = useMkrOnHat();
   const [voting, setVoting] = useState(false);
   const { data: votedProposals } = useVotedProposals();
-  const network = getNetwork();
   const bpi = useBreakpointIndex();
   const { comments } = useExecutiveComments(proposal.address);
-  const canVote = !!account;
```
adamgoth, line 39: I liked this because if the `canVote` logic changes, it just needs to be updated in one spot.

adamgoth (commented):
Looking good, nice work 💪

adamgoth (commented):
I'm noticing a console warning on this branch, `Warning: React has detected a change in the order of Hooks called by Header.`

modules/executive/components/ExecutiveOverviewCard.tsx
rafinskipg: yes? mh ok

rafinskipg: i think !account is something that is more or less extended through the app ( and in general on web3 apps),

modules/app/hooks/useAccount.ts
rafinskipg: fixed

modules/app/hooks/useVoteProxyOldAddress.ts
rafinskipg: fixed

modules/executive/components/ExecutiveOverviewCard.tsx
rafinskipg: restored

PR #310, opened 2022-01-25
convert voterWeight to BigNumber in useExecutiveComments hook

tyler17:
The UI showed NaN in the voter weight of an executive comment when the weight was >1000 because the db would return commas in the string representation of the numbers, which caused an error when converting them to BigNumbers.

This PR converts the voterWeight from a string to a bigNumber, minding the commas, in the useExecutiveComments hook.

Poll comments use type BigNumber for the voterWeight variable, so this reduces differences between the exec comments and poll comments types used by the components.

To test:
Pantera's vote weight is shown here: https://governance-portal-v2-fpiwhy0pj-dux-core-unit.vercel.app/executive/template-executive-vote-offboarding-usdt-collateral-core-unit-budget-distributions-october-1-2021?network=mainnet#comments
but not on production: https://vote.makerdao.com/executive/template-executive-vote-offboarding-usdt-collateral-core-unit-budget-distributions-october-1-2021?network=mainnet#comments

### DIFF
--- a/modules/comments/components/CommentItem.tsx
+++ b/modules/comments/components/CommentItem.tsx
@@ -7,7 +7,7 @@ import { getNetwork } from 'lib/maker';
 import useAccountsStore from 'modules/app/stores/accounts';
 import DelegateAvatarName from 'modules/delegates/components/DelegateAvatarName';
 import AddressIconBox from 'modules/address/components/AddressIconBox';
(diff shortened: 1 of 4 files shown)


adamgoth (approved):
LGTM. Is this an issue we could have with poll comments too, or just executive comments?

tyler17:
It doesn't affect polls because we don't get the voter weight from the comments database for the polls. Instead we use the voter weight from the vote breakdown from spock for that.  Whereas with executives we get the vote weight from the database.

Issue #353, opened 2022-02-11
Multiple vulns in vote.maker.dao X-XSS, clickjacking, Content header not se, cgi directories found 


philipjonsen:
Target IP:          76.76.21.21
+ Target Hostname:    vote.makerdao.com
+ Target Port:        80
+ Start Time:         2022-02-11 16:58:17 (GMT0)
---------------------------------------------------------------------------
+ Server: Vercel
+ The anti-clickjacking X-Frame-Options header is not present.
+ The X-XSS-Protection header is not defined. This header can hint to the user agent to protect against some forms of XSS
+ Uncommon header 'x-vercel-id' found, with contents: arn1::85jhw-1644595098595-f56427e1a8ad
+ Uncommon header 'refresh' found, with contents: 0;url=https://vote.makerdao.com/
+ The X-Content-Type-Options header is not set. This could allow the user agent to render the content of the site in a different fashion to the MIME type
+ Root page / redirects to: https://vote.makerdao.com/
+ No CGI Directories found (use '-C all' to force check all possible dirs)
+ Uncommon header 'x-vercel-error' found, with contents: BAD_REQUEST
+ Server: Vercel
+ The anti-clickjacking X-Frame-Options header is not present.
+ The X-XSS-Protection header is not defined. This header can hint to the user agent to protect against some forms of XSS
+ Uncommon header 'x-vercel-id' found, with contents: arn1::85jhw-1644595098595-f56427e1a8ad
+ Uncommon header 'refresh' found, with contents: 0;url=https://vote.makerdao.com/
+ The X-Content-Type-Options header is not set. This could allow the user agent to render the content of the site in a different fashion to the MIME type
+ Root page / redirects to: https://vote.makerdao.com/
+ No CGI Directories found (use '-C all' to force check all possible dirs)
+ Uncommon header 'x-vercel-error' found, with contents: BAD_REQUEST
- STATUS: Completed 560 requests (~8% complete, 4.0 minutes left): currently in plugin 'Site Files'
- STATUS: Running average: 100 requests: 0.02828 sec, 10 requests: 0.0278 sec.
V:Fri Feb 11 16:58:38 2022 - 308 for GET:       /vote.makerdao.com.cer
V:Fri Feb 11 16:58:38 2022 - 308 for GET:       /site.sql
V:Fri Feb 11 16:58:38 2022 - 308 for GET:       /site.sql
[...]

PR #355, opened 2022-02-14
Add a system to generate nonces for comments, and change security headers 

rafinskipg:
## Security headers: 
- Added the following headers based on this issue: https://github.com/makerdao/governance-portal-v2/issues/353
```
const securityHeaders = [
  // Adds x-xss-protection
  {
    key: 'X-XSS-Protection',
    value: '1; mode=block'
  },

  // adds x-frame-options
  {
    key: 'X-Frame-Options',
    value: 'SAMEORIGIN'
  },

  // adds x-content-type
  {
    key: 'X-Content-Type-Options',
    value: 'nosniff'
  }
];

```

## Poll Comments. 

Based on the security report, i have implemented:

- Before signing a message a "nonce" is requested from the API. This nonce is a combination of address + a random UUID that gets inserted on the database.

- The message has to be signed including that nonce.

- When posting the message the backend will check that: the nonce exists on the database for that address, that the signed message is actually the same nonce than the database, and then will delete the ocurrences.

- The voting weight for the executive comments is fetched on the backend so no one can fake it on the request


modules/comments/api/nonce.ts
```diff
@@ -0,0 +1,60 @@
...
+  const nonce = `/${address}/${uid}`;
+
+  // If we have a previous nonce, use that.
+  const previousNonces = await collection.find({
+    address
+  });
+
+  if (previousNonces.length > 0) {
+    return previousNonces[0].nonce;
+  }
```
adamgoth, line 23: You could do this check before lines 12-14

b-pmcg:
nice, looking good

b-pmcg (approved):
huzza

PR #357, opened 2022-02-16
Remove query parameters on api docs

rafinskipg:
Fix vulnerability query HTML injection can be done to swagger through query parameters. Force the user to access the route without query parameters. 

Added a middleware https://nextjs.org/docs/middleware that redirects any request with query parameters to non query parameters

### DIFF
--- a/pages/api-docs/index.tsx
+++ b/pages/api-docs/index.tsx
@@ -8,7 +8,6 @@ import Link from 'next/link';
 
 const ApiDoc = ({ spec }: InferGetStaticPropsType<typeof getStaticProps>) => {
   const router = useRouter();
-  console.log(router.query);
   return Object.keys(router.query).length > 0 ? (
     <Box>
       <h2>Invaid Route</h2>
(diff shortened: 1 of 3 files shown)


pages/api-docs/index.tsx
```diff
@@ -1,11 +1,23 @@
...
+import Link from 'next/link';
 
 const ApiDoc = ({ spec }: InferGetStaticPropsType<typeof getStaticProps>) => {
-  return <SwaggerUI spec={spec} />;
+  const router = useRouter();
+  return Object.keys(router.query).length > 0 ? (
+    <Box>
+      <h2>Invaid Route</h2>
+      <Link href="/api-docs">
+        <a title="See API docs">Go to the API docs.</a>
```
adamgoth, line 15: Is the `a` tag necessary?

adamgoth (approved):

pages/api-docs/index.tsx
rafinskipg: yes otherwise the title don't appear

PR #365, opened 2022-03-03
Update so abstain cannot be selected with other options

adamgoth:
### Link to Shortcut ticket:

N/A

### What does this PR do?

Updates ranked-choice component so that abstain cannot be combined with other options

### Steps for testing:

Go to a ranked-choice poll and try to select abstain with other options

Screen Shot 2022-03-03 at 1 59 51 PM
Screen Shot 2022-03-03 at 1 59 42 PM

This assumes `Abstain` will always be the first option in a ranked-choice poll. It will break if this is not the case.


rafinskipg (approved):
let's hope that is always the first option

adamgoth:
> let's hope that is always the first option

It's actually looking for "Abstain" text now instead of just taking the first option.

PR #369, opened 2022-03-07
Adding pagination

rafinskipg:
- We trying to fetch the spell data for the executives on the backend, because it was failing to load the executives page (big FE query that blocks the following requests)

We moved it to the backend, so the executives come with some of the spell data 

Why? 

Is used on the frontend by several components:
- sorting all the executives by MKR
- executive indicator uses all the proposals that have been scheduled 

The problem is that fetching all the spellData on the backend is blocking now the serverside rendering/static generation on the build, it takes a lot of time to do for all the executives.

This is a problem that will scalate over time, the more executives, the worse it gets.

## Short term solution:
- I'm adding a pagination to fetch 0-20 executives on the first load, and then allow the user to fetch more

## Better solution:
Our system should be able to aggregate that information asynchronously and the FE should be able to query by different states:

- Active executives, scheduled dates, and so on.

### DIFF
--- a/modules/delegates/api/fetchDelegates.ts
+++ b/modules/delegates/api/fetchDelegates.ts
@@ -99,7 +99,7 @@ export async function fetchDelegates(network?: SupportedNetworks): Promise<Deleg
   const delegatesInfo = await fetchDelegatesInformation(currentNetwork);
 
   const contracts = getContracts(networkNameToChainId(currentNetwork));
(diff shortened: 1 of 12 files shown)


adamgoth (approved):
LGTM, initial load is much quicker. One tradeoff is we lose the ability to sort all proposals by MKR support.

PR #375, opened 2022-03-10
Sort executives on the server side

rafinskipg:
- Pass sortBy to the backend
- Each executive MKR support is fetched and catched so the pagination can use that data

## Future:

In the future we should be able to query those values without having to recache data


adamgoth:
Noticing two things; filter text needs to be formatted from `mkr` to `MKR Support` and it looks like it's maybe sorting the wrong direction.

Screen Shot 2022-03-10 at 10 15 28 AM

PR #376, opened 2022-03-10
Feature/sc 1135/implement improved mkr liquidity panel

adamgoth:
### Link to Shortcut ticket:

https://app.shortcut.com/dux-makerdao/story/1135/implement-improved-mkr-liquidity-panel

### What does this PR do?

Adds uniswap v3 and compound liquidity to liquidity panel

Pools protocols together, adds expanding detail sections

### Steps for testing:

Go to exec page, see liquidity updates in sidebar ✨

### DIFF
--- a/modules/gql/queries/uniswapV3MkrSupply.ts
+++ b/modules/gql/queries/uniswapV3MkrSupply.ts
@@ -1,0 +1,10 @@
+import { gql } from 'graphql-request';
+
+export const uniswapV3MkrSupply = gql`
+  query uniswapV3MkrSupply($argMkrAddress: String!) {
+    token(id: $argMkrAddress) {
+      symbol
+      totalSupply
+    }
+  }
+`;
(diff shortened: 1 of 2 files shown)


b-pmcg:
Looks great, is opening both expandos at the same time intended behavior? Also on goerli the request never resolves and the loading animation goes on forever.

adamgoth:
> Looks great, is opening both expandos at the same time intended behavior? Also on goerli the request never resolves and the loading animation goes on forever.

Expanding/collapsing both was my intention. We could add separate state for each subgroup, but I don't see the point personally. What do you think?

Also I didn't think about goerli... we'd have to find subgraphs/APIs for those pools, if they exist..

b-pmcg:
> Also I didn't think about goerli... we'd have to find subgraphs/APIs for those pools, if they exist..

just meant showing 0 or "n/a" or something if it the request fails, but its probably fine

adamgoth:
> > Also I didn't think about goerli... we'd have to find subgraphs/APIs for those pools, if they exist..
> 
> just meant showing 0 or "n/a" or something if it the request fails, but its probably fine

Makes sense. Just pushed some updates so that it shows 0 on goerli, like the others.

PR #382, opened 2022-03-16
Fix/timeout api delegates

rafinskipg:
- Do not fetch all the vote history for delegates just to extract the last voted date
- Do not fetch MKR support for executives just to fetch the supporting executives

Reduced time loads on delegates api

### DIFF
--- a/modules/gql/queries/allCurrentVotes.ts
+++ b/modules/gql/queries/allCurrentVotes.ts
@@ -12,3 +12,16 @@ export const allCurrentVotes = gql`
     }
   }
 `;
+
+export const lastPollVote = gql`
+  query allCurrentVotes($argAddress: String!) {
+    allCurrentVotes(argAddress: $argAddress, first: 1) {
+      nodes {
+        pollId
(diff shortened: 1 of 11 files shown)


modules/gql/queries/allCurrentVotes.ts
```diff
@@ -12,3 +12,16 @@ export const allCurrentVotes = gql`
     }
   }
 `;
+
+export const lastPollVote = gql`
```
adamgoth, line 16: Maybe a separate file? We should either have all queries by module in the same file or have them all in separate files. But not mixed convention.

modules/gql/queries/allCurrentVotes.ts
```diff
@@ -12,3 +12,16 @@ export const allCurrentVotes = gql`
...
+  query allCurrentVotes($argAddress: String!) {
```
adamgoth, line 17: This query name should be updated to reflect what it's fetching.

adamgoth (approved):

PR #385, opened 2022-03-16
Refactor delegates api to sort by mkr, change home page data fetching

rafinskipg:
- Change Delegates API to sort by MKR or date

- Change home page data fetching mechanism to use SWR instead of that useEFect

- Add first steps of top delegates component


modules/delegates/api/fetchChainDelegates.ts
```diff
@@ -16,18 +17,20 @@ export async function fetchChainDelegates(
...
   const contracts = getContracts(chainId);
 
-  const mkrStaked = await Promise.all(
-    delegates.map(async delegate => {
+  const delegatesWithMkrStaked: DelegateContractInformation[] = await Promise.all(
+    delegates.map(async (delegate): Promise<DelegateContractInformation> => {
       // Get MKR delegated to each contract
       const mkr = await contracts.chief.deposits(delegate.voteDelegate);
-      return formatValue(mkr, 'wad', 18, false);
+      const delegationHistory = await fetchDelegationHistory(delegate.voteDelegate, network);
```
b-pmcg, line 24: With the new query by array we can fetch the history once and filter the results by the delegate we want.

rafinskipg: ok, so we query for all of them and then separate, got it!

PR #392, opened 2022-03-18
Create internal link component

adamgoth:
Create a reusable internal link component for consistent implementation


modules/executive/components/ExecutiveOverviewCard.tsx
```diff
@@ -178,35 +178,33 @@ export default function ExecutiveOverviewCard({
...
-            <Link
-              href={{ pathname: '/executive/[proposal-id]' }}
-              as={{ pathname: `/executive/${proposal.key}` }}
-              passHref
+            <InternalLink
+              href={'/executive/[proposal-id]'}
+              title="View executive details"
+              as={`/executive/${proposal.key}`}
             >
-              <ThemeUILink variant="nostyle" title="View Poll Details" sx={{ width: '100%' }}>
```
rafinskipg, line 196: does this sx={{ width: '100%' }}> that now is removed affect the UI?

rafinskipg (approved):

modules/executive/components/ExecutiveOverviewCard.tsx
```diff
@@ -81,17 +81,17 @@ export default function ExecutiveOverviewCard({
...
           <Box>
             <Flex sx={{ flexDirection: 'column', height: '200px' }}>
-              <Link
-                href={{ pathname: '/executive/[proposal-id]' }}
-                as={{ pathname: `/executive/${proposal.key}` }}
-                passHref
+              <InternalLink
+                href={'/executive/[proposal-id]'}
+                title="View executive details"
+                as={`/executive/${proposal.key}`}
```
b-pmcg, line 87: Nice, great idea. One note, I believe the `as` property is no longer required and we can pass the dynamic route straight into `href`. Check this out for more: https://nextjs.org/docs/api-reference/next/link

rafinskipg: yeah

adamgoth: > Nice, great idea. One note, I believe the `as` property is no longer required and we can pass the dynamic route straight into `href`. Check this out for more: https://nextjs.org/docs/api-reference/next/link

Sweet, thanks for the heads up. Nice thing is in the future if a change like that occurs again, we'll hopefully only have to update it in one place.

PR #395, opened 2022-03-21
Refactor/ballot context

rafinskipg:
- Removes ballot store
- Improves code by avoiding selecting the transaction in different places
- Improves code by delegating single responsability to the context instead of different hooks
- Resets ballot when changing network or account


adamgoth (approved):
Good stuff 💯

PR #399, opened 2022-03-22
Feat/merge dai UI theme

adamgoth:
### What does this PR do?

Merges dai UI theme into app's theme file. No longer need to import the theme package and overwrite it for changes.

### Steps for testing:

Click through out the app, make sure things look right


b-pmcg:
looks good so far, I tried playing around to see if i could fix some of the size issues, but the values seem to be correct. It appears to only affect the @reach components, some of the styles don't seem to be applying to them. I've seen that issue in the past, but it doesn't explain why it would suddenly stop working now.

adamgoth:
> looks good so far, I tried playing around to see if i could fix some of the size issues, but the values seem to be correct. It appears to only affect the @reach components, some of the styles don't seem to be applying to them. I've seen that issue in the past, but it doesn't explain why it would suddenly stop working now.

Took me a bit to figure out, but somehow had chopped off the primary button style when copying things over. A lot of other styles were based off that, so some stuff wasn't looking right. Should be fixed now. I still can't get the `theme` prop type for `ThemeProvider` to be happy though.

b-pmcg (approved):

adamgoth:
Going to merge this into the landing page redesign branch instead of develop as I think makes more sense with to combine it with all the changing designs from the new landing page

PR #417, opened 2022-04-06
Landing page - resources section

adamgoth:


modules/home/components/ResourcesLanding.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const CategoryButton = ({ label, color, active, onClick }) => {
+  return (
+    <Button
+      variant="outline"
+      onClick={onClick}
+      sx={{
+        background: active ? 'onPrimary' : 'rgba(255, 255, 255, 0.1)',
+        px: [2, 2, 3],
+        border: 'none'
```
rafinskipg, line 15: maybe add the title to the buttton too

modules/home/helpers/resources.ts
```diff
@@ -0,0 +1,119 @@
+export enum ResourceCategory {
```
rafinskipg, line 1: instead of in a helpers file i would put it under a component folder, because this info only affects to that component.

components/ResourcesLanding/resources.tsx

modules/home/helpers/resources.ts
```diff
@@ -0,0 +1,119 @@
...
+  ALL_RESOURCES = 'All Resources',
+  GOVERNANCE = 'Governance',
+  PRODUCTS_AND_TOOLS = 'Products & Tools',
+  DEVELOPERS = 'Developers'
+}
+
+enum ResourceBackground {
+  GOVERNANCE = '/assets/resource_card_green.png',
```
rafinskipg, line 9: Why don't we use CSS gradients instead of images?

rafinskipg (changes requested):
3 small comments, but looks really nice:
- Folder for the data 
- Gradients in case of images, is possible?

Small padding issue on desktop:
image

PR #421, opened 2022-04-12
Update tally caching

adamgoth:
### What does this PR do?

Instead of caching the raw tally result, instead cache the fully parsed tally response, if the poll voting period has ended.


b-pmcg (approved):
LGTM, I notice only one failing test so I pushed a quick fix for it. Hopefully this PR will solve a lot of our data fetching issues in the tests.

PR #426, opened 2022-04-14
feature/sc 1267/implement header footer

b-pmcg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1267/implement-header-footer

### What does this PR do?
adds new footer, header, menu

Header menu:
Screen Shot 2022-04-19 at 1 23 07 PM

New footer:
Screen Shot 2022-04-19 at 1 23 41 PM


modules/app/components/layout/Footer/LandingFooter.tsx
```diff
@@ -0,0 +1,228 @@
+import { Flex, Text, useColorMode } from 'theme-ui';
+import { Icon } from '@makerdao/dai-ui-icons';
```
rafinskipg, line 2: If it's only the footer for the homepage , maybe we can put it under "modules/home"?

rafinskipg (approved):

modules/app/components/layout/Footer/NewFooter.tsx
```diff
@@ -0,0 +1,235 @@
...
+      { title: 'GitHub', url: 'https://www.github.com/makerdao', icon: 'github' }
+    ],
+    makerdux: [
+      { title: 'Discord', url: 'https://discord.gg/GHcFMdKden', icon: 'discord' },
+      { title: 'Twitter', url: 'https://twitter.com/MakerDUX', icon: 'twitter' },
+      { title: 'GitHub', url: 'https://github.com/makerdao/governance-portal-v2', icon: 'github' },
+      { title: 'Canny', url: 'https://makergovernance.canny.io/', icon: 'canny' },
+      { title: 'Immunifi', url: 'https://immunefi.com/bounty/makerdao/', icon: 'immunifi' }
+    ]
+  };
```
adamgoth, line 156: This looks possibly duplicated in the `LandingFooter` component. Perhaps we can have a common file to help with all the links and logos?

modules/app/components/layout/Footer/NewFooter.tsx
```diff
@@ -0,0 +1,235 @@
...
+          // // top: 300,
+          // // pt: '500px',
+          // zIndex: -1,
+          // position: 'absolute',
+          // transform: 'scale(1.5)',
+          // // transform: 'rotate(21deg)',
+          // backgroundImage,
+          // // backgroundSize: '300%',
+          // backgroundRepeat: 'no-repeat',
+          // backgroundPosition: 'center bottom'
```
adamgoth, line 183: Still WIP? ✂️

modules/web3/hooks/useGasPrice.ts
```diff
@@ -0,0 +1,18 @@
...
+import { fetchGasPrice } from '../helpers/fetchGasPrice';
+
+type GasResponse = {
+  data?: number | string | undefined;
+  loading: boolean;
+  error?: Error;
+};
+
+export const useGasPrice = (): GasResponse => {
+  const { data, error } = useSWR('fetch-gas', () => fetchGasPrice('fast'));
```
adamgoth, line 11: Our default `refreshInterval` for SWR is set to 60 seconds, so if we want it any quicker than that, we should specify it here an as option. I think we could maybe increase to 10 or 15 seconds, but probably not a huge deal. Wdyt?

adamgoth (commented):
Looking good!

Left some minor comments. I think you may need to merge the latest changes from the main feature branch before merging because I notice you still have some old code included such as the system stats.

modules/app/components/layout/Footer/NewFooter.tsx
b-pmcg: ah weird, NewFooter should have been deleted, I renamed the file to LandingFooter. Nice catch

modules/web3/hooks/useGasPrice.ts
b-pmcg: I initially was thinking it should be on the slower side since most of our transactions don't really need to be gas competitive, but I'd be open to speeding it up. Another open question I meant to pose to everybody would be which speed setting we should default to. I chose 'fast' because it was kind of a middle ground.

https://www.etherchain.org/api/gasnow

PR #429, opened 2022-04-20
Tabs navigation

rafinskipg:
Adds sticky tabs navigation and smooth scroll on the homepage

image


adamgoth:
Screen Shot 2022-04-20 at 12 32 32 PM

Think it needs maybe a little more space between each to match design. I played around with it a little bit and it's nothing with the code but feels a little unnatural to me. Maybe the links should have some sort of hover state styling, what do you think?

PR #436, opened 2022-04-26
use new footer sitewide

b-pmcg:
### What does this PR do?

Uses the new Footer as the standard footer sitewide, removing references to the old Footers & `shortenFooter` property.


adamgoth (approved):
LGTM 💸

PR #445, opened 2022-05-02
Stop refetching spell data client side

adamgoth:
Removes code that is fetching spellData client-side that we already have access to from our server-side fetching


b-pmcg (approved):
lgtm, good catch

PR #446, opened 2022-05-02
use batchprovider for analyze spell

rafinskipg:
use batch provider for executives,


b-pmcg:
LGTM, since the batch provider handles batching on its own, maybe we should use it for all contracts. I made a PR for that here https://github.com/makerdao/governance-portal-v2/pull/447/files

PR #451, opened 2022-05-03
Fix/mkr locked delegate

b-pmcg:
### What does this PR do?

- Removes a call to fetch lock events from inside a loop (when fetching all delegates) because we can fetch the events all at once with an array. 
- Updates other areas that depended on the original implementation

~~Blocked by this update to gov polling db, which must be merged first~~ merged: https://github.com/makerdao/gov-polling-db/pull/107/files


adamgoth (approved):
LGTM 💸

PR #457, opened 2022-05-05
POC to filter active polls at the gql level

b-pmcg:
### What does this PR do?

Adds better support for using graphql filters with our requests. We can now create filters using the `ActivePollsRecordFilter` schema in (`/generated`) and pass to our graphql requests.
To add a new filter follow the pattern in `gqlFilters.ts` and add the type.

### DIFF
--- a/modules/gql/gqlFilters.ts
+++ b/modules/gql/gqlFilters.ts
@@ -1,0 +1,7 @@
+export const queryFilters = {
+  active: { filter: { endDate: { greaterThanOrEqualTo: Math.floor(Date.now() / 1000) } } }
+};
+export const getQueryFilter = (filterName: string | undefined) => {
+  if (!filterName) return null;
+  return queryFilters[filterName];
+};
--- a/modules/gql/gqlFilters.ts
+++ b/modules/gql/gqlFilters.ts
@@ -1,7 +1,33 @@
-export const queryFilters = {
-  active: { filter: { endDate: { greaterThanOrEqualTo: Math.floor(Date.now() / 1000) } } }
+import { QueryFilterNames } from './gql.constants';
+import { PollsQuery, QueryFilterList } from './types';
+
+export const queryFilters: QueryFilterList = {
+  active: () => ({
+    filter: { endDate: { greaterThanOrEqualTo: Math.floor(Date.now() / 1000) } }
+  }),
+  range: (start, end) => ({
+    filter: {
+      and: [{ pollId: { greaterThanOrEqualTo: start } }, { pollId: { lessThanOrEqualTo: end } }]
+    }
+  }),
+  pollId: pollId => ({ filter: { pollId: { equalTo: pollId } } })
 };
-export const getQueryFilter = (filterName: string | undefined) => {
-  if (!filterName) return null;
-  return queryFilters[filterName];
+
+export const getQueryFilter = (filterName: QueryFilterNames | undefined, ...args: any): PollsQuery | null => {
+  if (filterName && queryFilters[filterName]) {
+    switch (filterName) {
+      case QueryFilterNames.Active:
+        return queryFilters.active();
+
+      case QueryFilterNames.Range:
+        return queryFilters.range(args[0], args[1]);
+
+      case QueryFilterNames.PollId:
+        return queryFilters.pollId(args[0]);
+
+      default:
+        break;
+    }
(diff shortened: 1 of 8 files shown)


b-pmcg:
@adamgoth this is a very inelegant implementation, but it might be a good solution for streamlining the data we fetch for the active poll count.

adamgoth (approved):
Seems to work well. I'm wondering if it makes sense to generalize the `getPolls` function to do the caching and query there rather than writing separate functions like `_getFilteredPolls` and `_getAllPolls`. Seems more scalable if we end up adding more filters to it. But in any case, this is a nice improvement I think.

b-pmcg:
> Seems to work well. I'm wondering if it makes sense to generalize the `getPolls` function to do the caching and query there rather than writing separate functions like `_getFilteredPolls` and `_getAllPolls`. Seems more scalable if we end up adding more filters to it. But in any case, this is a nice improvement I think.

yeah definitely, this was kind of rushed just to get it working. One thing I'll look into is how caching would work if we are doing filtering in the same query. Because if we cache everything we return everything in the cache. I guess that would be fine because the cache should be returned instantly, but then we would have to filter the results again to get the subset we're looking for.

modules/gql/gqlFilters.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      case QueryFilterNames.PollId:
+        return queryFilters.pollId(args[0]);
+
+      default:
+        break;
+    }
+  }
+  return null;
+};
```
b-pmcg, line 33: @adamgoth This is a bit redundant, initially I was trying to get the following to work by using generic types:

```
export const getQueryFilter = (filterName: QueryFilterNames | undefined, ...args: any): PollsQuery | null => {
  return filterName && queryFilters[filterName] ? queryFilters[filterName](...args) : null;
};
```

But without explicitly passing the parameters to each method, I was getting a typescript error complaining that `a spread argument must either have a tuple type or be passed to a rest parameter`.

b-pmcg:
@adamgoth if you think it's ready feel free to merge it into the feature branch. In addition to the `active` filter, I added `pollId` & `range` filters. The range filter isn't implemented anywhere yet, but I tested that it works, we might be able to replace the way we're currently doing it in `fetchPolls.ts` ln 134

modules/polling/api/fetchPollBy.ts
```diff
@@ -1,51 +1,54 @@
...
 
 export async function fetchSpockPollById(
   pollId: number,
   network: SupportedNetworks
 ): Promise<PollSpock | undefined> {
-  const polls = await fetchAllSpockPolls(network);
-  return polls.find(poll => poll.pollId === pollId);
+  const filter = getQueryFilter(QueryFilterNames.PollId, pollId);
+  const [poll] = await fetchSpockPolls(network, filter);
+  return poll;
```
rafinskipg, line 22: what is the rationale to remove a query that can filter by multihash and now querying al the polls again?

rafinskipg (changes requested):
I have seen several things that we need to fix. 

First: the API endpoint returns the total aggregated results of active polls. Those are numbers that are used on the UI and filters, to let know the user what results are expected.

image

In order to have this data we need that the spock query actually returns the totals, (IGNORING THE FILTER). Or to have a new query on spock that just counts totals. 

I suggest we alter the spock response to include those totals. This could be some standard thing accross all queries with filters.

Second, i see we removed the filterbymultihash query and instead now we are iterating over all the polls. isn't it better to just query one poll?

modules/polling/api/fetchPollBy.ts
b-pmcg: @rafinskipg I'm seeing two different things here. The query added above here `const [poll] = await fetchSpockPolls(network, filter);` fetches only one poll using the poll id. It uses the `PollId` query filter. The filter works exactly the same way as the multihash filter below where we do:
```
  variables: {
      argPollMultihash: `${slug}%`
    }
```
The only difference is I've moved the filters to their own file and defined it explicitly.

Below I'm not sure why the code is commented out, but I suggest we use the new filter pattern I added for gql queries. It would just be a matter of moving the `argPollMultihash: `${slug}%` variable into the gqlFilters file.

b-pmcg:
closing now in favor of #467

PR #462, opened 2022-05-07
Updates and XXS fix.

philipjonsen:
### Link to Shortcut ticket:
Updates and XXS fix.


rafinskipg:
@philipjonsen Seems like there's a problem with the build https://github.com/swagger-api/swagger-ui/issues/7970

PR #469, opened 2022-05-12
filter comments by txHash when fetching them instead of when saving them

rafinskipg:
We were checking that the txHash was correct when saving a comment, but since some people can put a really slow tx, we might want to verify the txHash when fetching the comments. 

We should test if this is enough or we need to do more changes

Comments are still saved on txPending


modules/comments/api/getPollComments.ts
```diff
@@ -51,6 +51,7 @@ export async function getPollComments(
     return {
       comment,
       isValid,
+      completed: i === 0 ? false : transaction && transaction.confirmations > 10,
```
b-pmcg, line 54: Awesome, looks really good. Curious about this line though, why is `i === 0 ? false`?

PR #470, opened 2022-05-17
Feat/polling gql pagination

b-pmcg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1005/add-sorting-for-polling-page

### What does this PR do?
enable cursor and pagination variables for `allWhitelistedPolls` query

example of how this works:
```js
// Get the next five active polls
const filter = getQueryFilter(QueryFilterNames.Active, { endDate: new Date().getTime() });

const queryVars = { ...filter, first: 5, after: poll.cursor };

const nextFivePolls = await getPolls(undefined, network, queryVars);
```


pages/polling/[poll-hash].tsx
```diff
@@ -64,6 +69,32 @@ const PollView = ({ poll }: { poll: Poll }) => {
   const { tally } = usePollTally(poll.pollId, 60000);
   const { comments, error: errorComments } = usePollComments(poll.pollId);
 
+  useEffect(() => {
+    const fetchFilteredPolls = async categories => {
+      const { polls: pollData } = await fetchJson(
```
Comment (rafinskipg) [inline pages/polling/[poll-hash].tsx:74]: i suggest we move this fetchJSON to a useSWR hook, so we can avoid refetching (it will use swr internal cache) if the query filter does not change.

modules/polling/api/fetchPolls.ts
```diff
@@ -68,26 +68,30 @@ export async function fetchAllPollsMetadata(pollList: PollSpock[]): Promise<Poll
...
   };
 
   const data = await gqlRequest<GqlQuery>(requestData);
 
-  const pollList: Maybe<ActivePollsRecord>[] = data.activePolls.nodes;
-  //TODO: maybe we should use the GQL generated types when applicable
+  const pollEdges: ActivePollEdge[] = data.activePolls.edges;
+
+  const pollList = pollEdges.map(({ node: poll, cursor }) => {
+    return { ...poll, cursor };
```
rafinskipg, line 84: i don't see where this cursor is used

rafinskipg (changes requested):

modules/polling/api/fetchPolls.ts
b-pmcg: we can't do serverside pagination right now because some of the filters are done in the client side. Soon I hope we can look into adding the categories to spock and then we can do filtering and pagination in the gql request. I left the cursor in there so it's available to use. We can still do SS pagination on the main `/polling` page.

pages/polling/[poll-hash].tsx
Comment (b-pmcg) [reply pages/polling/[poll-hash].tsx]: Good call, I added the hook.

adamgoth (approved):
Nice, this works well! We'll need to reconcile these updates with the new poll filter updates I worked on (still open in a separate PR: https://github.com/makerdao/governance-portal-v2/pull/470).

rafinskipg (approved):

PR #478, opened 2022-05-20
add tooltip to account modal and use 2 decimal places for vote weight…

b-pmcg:
Adds the poll vote weight tooltip to the account box

Screen Shot 2022-05-20 at 5 11 22 PM

### DIFF
--- a/modules/app/components/layout/header/VotingWeight.tsx
+++ b/modules/app/components/layout/header/VotingWeight.tsx
@@ -1,8 +1,11 @@
-import { Flex, Text } from 'theme-ui';
+import { Box, Flex, Text } from 'theme-ui';
+import { Icon } from '@makerdao/dai-ui-icons';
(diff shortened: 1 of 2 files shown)


adamgoth:
Looks good to me. But taking a look at the screenshot, I wonder if we should update the tooltip to reflect that the weight displayed for delegates is the balance of MKR delegated to them. 

If a user is a delegate, the balance in their wallet isn't really relevant since all voting occurs as a delegate using the delegated balance.

adamgoth (approved):

rafinskipg:
> Looks good to me. But taking a look at the screenshot, I wonder if we should update the tooltip to reflect that the weight displayed for delegates is the balance of MKR delegated to them.
> 
> If a user is a delegate, the balance in their wallet isn't really relevant since all voting occurs as a delegate using the delegated balance.

this is a good idea

rafinskipg (approved):
nice. I also second @adamgoth 's comment

PR #484, opened 2022-05-25
Use filtered polls from store instead

adamgoth:
Instead of using API to refetch polls from poll detail page based on filter state, just use the filtered poll data from polling page prior to navigation to poll detail page


rafinskipg (approved):
makes sense to restore the client side filtering since we are still not capable of doing the "next" "prev" filters on the backend. 

But eventually we will need to call to the backend to get the correct next/prev

PR #486, opened 2022-05-26
add option tooltip to statsbox and include tooltips for account/addre…

b-pmcg:
### What does this PR do?
- Adds an optional tooltip prop to the StatsBox
- Passes a voting weight breakdown tooltip to account/address stats box

Screen Shot 2022-05-26 at 3 15 14 PM


tyler17 (approved):
Looks good 👍

PR #487, opened 2022-05-31
Store ballot in localstorage

rafinskipg:
### What does this PR do?
Stores ballot in local storage. Allows to perform page refreshes and keep the ballot there for the user to recover.
- Stored with the following keys: network, address. So if the network or address of the user change we pick another item from the storage

https://www.loom.com/share/1da42b0451734503a8d26d12cd56bc4a


adamgoth (approved):
LGTM 💸

PR #495, opened 2022-06-07
Fix wrong rpc argument

adamgoth:
### What does this PR do?

Pass chain id instead of provider name


b-pmcg:
Nice catch, how did you notice this? I wonder if this will fix the errors with ethers rate limit exceeded, or even some of our alchemy 429's. If ethers didn't know how to parse the RPC name string it would just return a provider with the default (rate-limited) keys. How great would it be if this fixes our provider issues.

b-pmcg:
Nice catch, how did you notice this? I wonder if this will fix the errors with ethers rate limit exceeded, or even some of our alchemy 429's. If ethers didn't know how to parse the RPC name string it would just return a provider with the default (rate-limited) keys. How great would it be if this fixes our provider issues.

PR #500, opened 2022-06-07
Ranked choice bug fix

tyler17:
Properly encodes ranked choice votes (including reversing the options array)


modules/polling/context/BallotContext.tsx
```diff
@@ -178,9 +179,19 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
...
+        byteArray[byteArray.length - i - 1] = optionIndex;
+      });
+      return fromBuffer(byteArray, {}).toString();
+    });
+
     const voteTxCreator = voteDelegateContract
-      ? () => voteDelegateContract['votePoll(uint256[],uint256[])'](pollIds, pollOptions)
-      : () => polling['vote(uint256[],uint256[])'](pollIds, pollOptions);
+      ? () => voteDelegateContract['votePoll(uint256[],uint256[])'](pollIds, optionIds)
+      : () => polling['vote(uint256[],uint256[])'](pollIds, optionIds);
```
b-pmcg, line 194: Comparing it to how it used to be done in dai.js, it looks like we check the length of the options array to choose which `vote` method to use. Does it matter if we always use the one with `uint265[]`?

https://github.com/makerdao/dai.js/blob/dev/packages/dai-plugin-governance/src/GovPollingService.ts#L60-L66

b-pmcg (approved):
LGTM, as long as we're confident using the array vote method works when voting a single poll. I don't see why not as long as the parameters are passed as an array.

modules/polling/context/BallotContext.tsx
tyler17: Nice, yeah I tested it out with one poll vote, and it worked fine. The only benefit it seems is that it's *slightly* less gas if you're only voting on one poll, but it's only like a 5% gas price reduction so it doesn't seem worth the additional complication

rafinskipg (dismissed):
Code seems fine, but i would like to see unit tests that cover:
- more than 10 options (as found by the bignumber bug)
- 1 option
- 0 options

We need some code in place to make sure we dont remove the encoding again

@tyler17

b-pmcg (approved):
lgtm

PR #503, opened 2022-06-08
allow for 3 github tokens

tyler17:
Default to first token if 2nd or 3rd token not provided

### DIFF
--- a/modules/delegates/api/fetchGithubDelegates.ts
+++ b/modules/delegates/api/fetchGithubDelegates.ts
@@ -13,7 +13,7 @@ async function extractGithubInformation(
   folder: GithubPage
 ): Promise<DelegateRepoInformation | undefined> {
   try {
-    const folderContents = await fetchGitHubPage(owner, repo, folder.path);
+    const folderContents = await fetchGitHubPage(owner, repo, folder.path, 1);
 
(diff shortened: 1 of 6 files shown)


modules/delegates/api/fetchGithubDelegates.ts
```diff
@@ -13,7 +13,7 @@ async function extractGithubInformation(
   folder: GithubPage
 ): Promise<DelegateRepoInformation | undefined> {
   try {
-    const folderContents = await fetchGitHubPage(owner, repo, folder.path);
+    const folderContents = await fetchGitHubPage(owner, repo, folder.path, 1);
```
b-pmcg, line 16: Nice, think we could put the ID's into an enum? Might be easier to read if it's something like `GithubKeys.Delegates` or something similar.

dux-user: Sounds good 👍

tyler17: Good suggestion, I added the enum, and realized we should just use 3 different tokens, there's two different functions where we fetch the delegates folder, so just using the same token for both of those @b-pmcg

adamgoth (approved):
Good call, this should help reduce the chance of rate-limiting

rafinskipg (approved):

PR #520, opened 2022-06-20
fix poll parameters filters

rafinskipg:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1351/updates-to-poll-filters
### What does this PR do?

Restore poll type filter with the new poll metadata
image


tyler17:
> LGTM. Noticed we have Majority as an option but no polls with that value type. Should we exclude it if there are 0 results from all polls?

That makes sense to me

PR #521, opened 2022-06-21
include redis as a top level cache instead of filesystem

rafinskipg:
We include ioredis as a top level cache, based on the example described in this blog post https://dev.to/nuro/using-github-as-a-cms-2n28

We replaced the fsCacheGet / fsCacheSet methods with a `cacheGet` / `cacheSet` methods.

Now the cache module decides if using redis (if the user configured REDIS_URL environment variable) and defaults to filesystem cache if REDIS_URL is not configured. 

USE_FS_CACHE has been replaced by USE_CACHE

**Redis cloud**
- I started trying with Upstash, but we inmediately hit the size limit (1MB) per storage. Since we are storing a lot of data with the polls we need to go to a paid tier (480 per month), so i  created a redis on google cloud, but the estimated cost is $805.92/month 
image


lib/cache.ts
```diff
@@ -0,0 +1,115 @@
...
+): Promise<any> => {
+  const isRedisCache = !!config.REDIS_URL;
+
+  try {
+    const currentNetwork = network || DEFAULT_NETWORK.network;
+    const path = getFilePath(name, currentNetwork);
+
+    if (isRedisCache && redis) {
+      // Get redis data if it exists
+      const cachedData = await redis.get(path);
```
adamgoth, line 44: What happens here if `path` doesn't exist? Does it return null or throw an error?

adamgoth (approved):
LGTM 💸  just had one question about when `redis.get(path)` doesn't find any cache. I assume it returns null?

adamgoth (commented):
Looks like we need to update the polling test for cached items. I think it's related to the cache key name.

lib/cache.ts
rafinskipg: yes, returns null

PR #524, opened 2022-06-22
Remove default provider from getENS

adamgoth:
Significantly reduce the number of calls from ethers by passing getENS the connected provider library instead of using ethers default provider 

Cleans up Davatar, removes default provider usage

Prevents gas price fetching not on mainnet


b-pmcg (approved):
LGTM, if it's reducing calls let's ship it

PR #525, opened 2022-06-23
Update logger implementation

rafinskipg:
Replaces console.logs with formatted log messages and adds functions names so we know where the errors come from : 

image
image

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -7,6 +7,7 @@ import { CurrencyObject } from 'modules/app/types/currency';
 import { hexZeroPad, stripZeros } from 'ethers/lib/utils';
 
 import round from 'lodash/round';
+import logger from './logger';
 
 export function bigNumberKFormat(num: CurrencyObject): string {
   invariant(num && num.symbol && num.toBigNumber, 'bigNumberKFormat must recieve a maker currency object');
@@ -215,9 +216,7 @@ const windowOpen = (
           onClose(shareDialog);
         }
       } catch (e) {
-        /* eslint-disable no-console */
-        console.error(e);
-        /* eslint-enable no-console */
+        logger.error('Window open', e);
       }
     }, 1000);
   }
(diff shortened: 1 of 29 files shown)


lib/utils.ts
```diff
@@ -215,9 +216,7 @@ const windowOpen = (
           onClose(shareDialog);
         }
       } catch (e) {
-        /* eslint-disable no-console */
-        console.error(e);
-        /* eslint-enable no-console */
+        logger.error('Window open', e);
```
adamgoth, line 219: Nit pick but maybe match function names and casing so it's easy to search and find in code base

So `windowOpen:` instead of `Window open`

modules/web3/hooks/useInactiveListener.ts
```diff
@@ -12,15 +13,15 @@ export function useInactiveListener(suppress = false) {
...
+          logger.error('Activate Web3 Connector: Failed to activate after chain changed', error);
         });
       };
 
       const handleAccountsChanged = (accounts: string[]) => {
         if (accounts.length > 0) {
           // eat errors
           activate(injectedConnector, undefined, true).catch(error => {
-            console.error('Failed to activate after accounts changed', error);
+            logger.error('Activate Web3 Connector: Failed to activate after accounts changed', error);
```
adamgoth, line 24: Maybe `handleAccountsChanged` instead?

modules/web3/hooks/useInactiveListener.ts
```diff
@@ -12,15 +13,15 @@ export function useInactiveListener(suppress = false) {
       const handleChainChanged = () => {
         // eat errors
         activate(injectedConnector, undefined, true).catch(error => {
-          console.error('Failed to activate after chain changed', error);
+          logger.error('Activate Web3 Connector: Failed to activate after chain changed', error);
```
adamgoth, line 16: Maybe `handleChainChanged` instead?

adamgoth (approved):
Left some nit-pick comments but this is a nice update 👍

PR #527, opened 2022-06-23
Fix/format mkr balance

adamgoth:
Use formatting function for formatting amounts of MKR

Remove decimals for zero values of MKR

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -239,3 +239,9 @@ export function openWindowWithUrl(url: string): void {
 
   windowOpen(url, windowConfig);
 }
+
+export function formatMkrAmount(mkrAmount: number | string): string {
+  return new BigNumber(mkrAmount).lte(0.01)
+    ? '≈0.00'
+    : new BigNumber(mkrAmount).toFormat(new BigNumber(mkrAmount).gt(999) ? 0 : 2);
+}
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -238,9 +238,3 @@ export function openWindowWithUrl(url: string): void {
 
   windowOpen(url, windowConfig);
 }
-
-export function formatMkrAmount(mkrAmount: number | string): string {
-  return new BigNumber(mkrAmount).lte(0.01)
-    ? '≈0.00'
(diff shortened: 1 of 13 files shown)


modules/delegates/components/DelegateMKRDelegatedStats.tsx
```diff
@@ -50,7 +50,7 @@ export function DelegateMKRDelegatedStats({
         label={'Total Active Delegators'}
       />
       <StatBox
-        value={typeof mkrStaked !== 'undefined' ? formatValue(mkrStaked) : '0.00'}
+        value={typeof mkrStaked !== 'undefined' ? formatValue(mkrStaked) : '0'}
```
rafinskipg, line 53: Don't we use here the formatMkr amount?

modules/delegates/components/DelegateOverviewCard.tsx
```diff
@@ -213,7 +213,7 @@ export function DelegateOverviewCard({ delegate }: PropTypes): React.ReactElemen
                     sx={{ fontSize: [3, 5], textAlign: ['left', 'right'] }}
                     data-testid="total-mkr-delegated"
                   >
-                    {totalStaked && totalStaked.gt(0) ? formatValue(totalStaked) : '0.00'}
+                    {totalStaked && totalStaked.gt(0) ? formatValue(totalStaked) : '0'}
                   </Text>
```
rafinskipg, line 217: here too, why do we use formatValue vs formatMKRAmount?

lib/utils.ts
```diff
@@ -239,3 +239,9 @@ export function openWindowWithUrl(url: string): void {
 
   windowOpen(url, windowConfig);
 }
+
+export function formatMkrAmount(mkrAmount: number | string): string {
```
rafinskipg, line 243: We have formatValue and formatMKRAmount and i'm starting to be confused, shouldn't we just use one function to format numbers through the app? 

In case we need 2 i asume is because this one rounds "near zero" values to ≈0.00 , shouldn't that be a parameter of the other format function? idk

adamgoth: You're right, I didn't realize `formatValue` handled the `>999` logic as well. So we can refactor to do the `<0.01` logic too.

rafinskipg (approved):

PR #528, opened 2022-06-24
Delegate Expiry: Delegated to expired banner

rafinskipg:
### What does this PR do?
Includes the hook to know if someone is delegatign to a expired delegate.

Modifies the delegatedTo api call to return the expiration date of the delegate
```
{
	"delegatedTo": [{
		"address": "0x84b05b0a30b6ae620f393d1037f217e607ad1b96",
		"expirationDate": "2023-02-28T19:07:12.000Z",
		"lockAmount": "4.87398",
		"events": [{
			"lockAmount": "4.873980000000000000",
			"blockTimestamp": "2022-03-24T16:30:12+00:00",
			"hash": "0x928919522d199d1171bcbd90664a67551743abbb000bbc48c9d3a87365a01988"
		}]
	}],
	"totalDelegated": 4.87398
}
```

Figma link: https://www.figma.com/file/k8mx0Fups2truCXvAZrqd0/Iterations?node-id=0%3A1

### DIFF
--- a/modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
+++ b/modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
@@ -1,0 +1,7 @@
+import { useAccount } from "modules/app/hooks/useAccount";
+
+export default function useDelegatedToExpired() {
+    const { account  } = useAccount() ;
+
+    
+}
--- a/modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
+++ b/modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
@@ -1,7 +1,18 @@
-import { useAccount } from "modules/app/hooks/useAccount";
+import { isBefore, sub } from 'date-fns';
+import useDelegatedTo from 'modules/delegates/hooks/useDelegatedTo';
+import { useActiveWeb3React } from 'modules/web3/hooks/useActiveWeb3React';
 
-export default function useDelegatedToExpired() {
-    const { account  } = useAccount() ;
+export default function useDelegatedToExpired(): boolean {
+  const { account: address, network } = useActiveWeb3React();
 
-    
+  const { data: delegatedToData } = useDelegatedTo(address, network);
+
+  const isADelegateAboutToExpire = delegatedToData.delegatedTo.reduce((prev, next) => {
+    // check if it has less than one week to expire or is expired
+    const isExpiredOrAboutToExpire = isBefore(new Date(next.expirationDate), sub(new Date(), { days: 7 }));
+
+    return prev || isExpiredOrAboutToExpire;
+  }, false);
+
+  return isADelegateAboutToExpire;
 }
--- a/modules/address/components/AddressDetail.tsx
+++ b/modules/address/components/AddressDetail.tsx
@@ -7,7 +7,6 @@ import useSWR from 'swr';
 import { fetchJson } from 'lib/fetchJson';
 import LastVoted from 'modules/polling/components/LastVoted';
 import AddressDelegatedTo from './AddressDelegatedTo';
-import { MKRDelegatedToAPIResponse } from 'pages/api/address/[address]/delegated-to';
 import SkeletonThemed from 'modules/app/components/SkeletonThemed';
 import { AddressMKRDelegatedStats } from './AddressMKRDelegatedStats';
 import AddressIconBox from './AddressIconBox';
@@ -15,6 +14,7 @@ import { useActiveWeb3React } from 'modules/web3/hooks/useActiveWeb3React';
 import { ErrorBoundary } from 'modules/app/components/ErrorBoundary';
 import AccountComments from 'modules/comments/components/AccountComments';
 import Tabs from 'modules/app/components/Tabs';
+import useDelegatedTo from 'modules/delegates/hooks/useDelegatedTo';
 
 type PropTypes = {
   address: string;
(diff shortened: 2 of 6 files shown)


lib/theme/colors.js
```diff
@@ -35,8 +35,8 @@ export const colors = {
   noticeAlt: '#FFFBEF',
   onNotice: '#EB7720',
   warning: '#CB532D',
-  warningAlt: '#FFF6F3',
-  onWarning: '#CB532D',
```
adamgoth, line 39: How come we're changing this from a red to white? Also looks like it's the same color for light and dark mode? Doesn't seem right.

modules/delegates/components/DelegatePicture.tsx
```diff
@@ -158,7 +158,7 @@ export function DelegatePicture({
                   position: 'absolute',
                   bottom: width / -12,
                   right: width / -7,
-                  size: width / 2.5,
+                  width: `${Math.round(width / 2.5)}px`,
```
adamgoth, line 161: Just wondering why we need this update?

modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
```diff
@@ -0,0 +1,23 @@
+import { add, isBefore } from 'date-fns';
+import useDelegatedTo from 'modules/delegates/hooks/useDelegatedTo';
+import { useActiveWeb3React } from 'modules/web3/hooks/useActiveWeb3React';
+
+export default function useDelegatedToExpired(): boolean {
+  const { account: address, network } = useActiveWeb3React();
+
+  const { data: delegatedToData } = useDelegatedTo(address, network);
```
adamgoth, line 8: Would it make more sense to add this check on about to expire to the `delegated-to` endpoint and then it would be part of the `useDelegatedTo` hook too and then just use that everywhere? It's just adding one more properly from reducing an array. But then it'd get cached with the API endpoint and SWR as well.

modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  const isADelegateAboutToExpire = delegatedToData
+    ? delegatedToData.delegatedTo.reduce((prev, next) => {
+        // check if it has less than one week to expire or is expired
+        const isExpiredOrAboutToExpire = isBefore(
+          new Date(next.expirationDate),
+          add(new Date(), { days: 7 })
```
adamgoth, line 15: Here we have 7 days but on the `AddressDelegatedTo` we use 30 days for about to expire. We should figure out which one to use, and then maybe make a constant in the app we can use so that if we update it, we just need to change the days in one place.

pages/_app.tsx
```diff
@@ -27,83 +27,99 @@ import debug from 'debug';
...
+                  <Banner
+                    variant="warning"
+                    content={
+                      <Text sx={{ fontWeight: '400' }}>
+                        You delegated MKR to a contract that is going to expire or has expired. Please migrate
+                        your MKR to a renewed Delegate Contract
+                      </Text>
+                    }
+                  />
+                )}
```
adamgoth, line 94: Maybe a separate component for this even though it's small, so we don't clutter up `_app.tsx` with copy and styles and what not.

pages/_app.tsx
```diff
@@ -27,83 +27,99 @@ import debug from 'debug';
...
+                      MozOsxFontSmoothing: 'grayscale'
+                    }
+                  }}
+                />
+                {activeBannerContent && <Banner content={activeBannerContent.content} />}
+                {hasDelegateExpiring && (
+                  <Banner
+                    variant="warning"
+                    content={
+                      <Text sx={{ fontWeight: '400' }}>
```
adamgoth, line 88: Isn't 400 the default font weight?

modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useActiveWeb3React } from 'modules/web3/hooks/useActiveWeb3React';
+
+export default function useDelegatedToExpired(): boolean {
+  const { account: address, network } = useActiveWeb3React();
+
+  const { data: delegatedToData } = useDelegatedTo(address, network);
+
+  const isADelegateAboutToExpire = delegatedToData
+    ? delegatedToData.delegatedTo.reduce((prev, next) => {
+        // check if it has less than one week to expire or is expired
```
adamgoth, line 12: We need to distinguish between `expired` or `expiring soon` so we can use that logic throughout the app for the banner and different flows.

modules/delegates/components/DelegatePicture.tsx
rafinskipg: I tried modifying the size of the delegate avacatr, and without this, the SVG would go crazy, if you put delegate picture to 20px , then the SVG goes to 500px, with this formula it works better

pages/_app.tsx
rafinskipg: not for the banner :S it goes with a bold font weight by dfefault

modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
rafinskipg: I thought the banner gives a sense of more urgency, and in the list of addresses delegating like you have more time to visualize that the problem is coming. We should ask deniz

modules/delegation-renewal/hooks/useDelegatedToExpired.tsx
rafinskipg: idk, seems like a really frontend logic to me, the api is giving us a expiryDate, that should be enough, having also "isAboutToExpire" or "isExpired" seems redundant information

pages/_app.tsx
adamgoth: This is the only use of it, so we can update it in `Banner` component I think. Otherwise we'll end up with mismatched banner styles eventually.

PR #533, opened 2022-06-28
create a wrapper for bignumber.js, add 'no exponent' config option

b-pmcg:
### What does this PR do?

- Clone bignumber and add a global config option to prevent exponent notation
- adds a named export `BigNumberJS` to alleviate confusion with ethers `BigNumber`
- adds test to prevent files from importing BigNumber directly from 'bignumber.js' package

### Steps for testing:
Check vote breakdown page on polls and make sure it doesn't break. This is the one that errored this morning: `https://<previewURL>/polling/Qmbs5Kx5`

https://github.com/MikeMcl/bignumber.js/issues/315#issuecomment-1020673911

### DIFF
--- a/lib/__tests__/bigNumber.spec.js
+++ b/lib/__tests__/bigNumber.spec.js
@@ -1,0 +1,30 @@
+import glob from 'glob';
+import path from 'path';
+import fs from 'fs';
+
+it('ensure bignumber.js is not used directly', () => {
+  const cwd = process.cwd();
+  const files = glob.sync('**/*', {
+    cwd,
+    ignore: [
+      'lib/bigNumberJs.ts', //ignore the file that wraps bignumber
+      'node_modules/**',
+      'dist/**',
+      '.github/**',
+      '.husky/**',
+      '.next/**',
+      '.yalc/**',
+      'cache/**',
+      'coverage/**',
+      'cypress/**'
+    ],
+    nodir: true
+  });
+  const filesWithBigNumber = files.filter(f => {
+    const extension = path.extname(f);
+    const data = fs.readFileSync(f, 'utf8');
+    if (['.js', '.ts'].includes(extension) && data.includes('bignumber.js')) return f;
+  });
+  //the only non-ignored file that should contain bignumber.js is this file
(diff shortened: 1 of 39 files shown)


tyler17 (approved):

tyler17:
LGTM.  I also tested out submitting a ranked choice ballot with 12 options selected, which also used to cause issues with exponential notation.

Wonder if we could add a test that fails if any file other than `lib/bigNumberJs.ts` tries to import bignumber.js

adamgoth (dismissed):
LGTM. One suggestion, not a big deal, but what if we export it as `BigNumberJs` and then always import it as that. That way we don't have a mix of sometimes `BigNumber` is the ethers version and sometimes it's the JS version. You'd just know right away when you see it.

b-pmcg:
> LGTM. I also tested out submitting a ranked choice ballot with 12 options selected, which also used to cause issues with exponential notation.
> 
> Wonder if we could add a test that fails if any file other than `lib/bigNumberJs.ts` tries to import bignumber.js

interesting, how would we do that, with Jest? Or maybe a custom linter rule?

b-pmcg:
> 

That makes sense, so in that case we wouldn't have a default export, right? Just a named export. This would work well with @tyler17's suggestion above.

tyler17:
Not sure, but custom linter rule seems like a good way to do it, I can look into that

tyler17:
was easier to just write a test that reads all the files, what do you think? https://github.com/makerdao/governance-portal-v2/pull/541

b-pmcg:
Oops, I forgot to quote @adamgoth in my comment above. I started implementing this but I realized there are too many places where we use a class called `BigNumber` to make a quick job. However I think it's a good idea so I changed the named export to `BigNumberJS` and in the files where we were already aliasing `BigNumber` as `BigNumberJS`, I just changed the import. Hopefully that's a good compromise and we can slowly phase out importing it as a default export over time.

b-pmcg:
The funny thing is, ethers actually uses BN.js under the hood, but calls it BigNumber. I guess it's far too late to change that now :P

rafinskipg (approved):

PR #536, opened 2022-06-28
Cache executive supporters 10 minutes

rafinskipg:
Adds cache to the executive supporters api endpoint for 10 minutes

### DIFF
--- a/pages/api/executive/supporters.ts
+++ b/pages/api/executive/supporters.ts
@@ -7,12 +7,24 @@ import { isSupportedNetwork } from 'modules/web3/helpers/networks';
 import { getContracts } from 'modules/web3/helpers/getContracts';
 import { networkNameToChainId } from 'modules/web3/helpers/chain';
 import { fetchExecutiveVoteTally } from 'modules/executive/api/fetchExecutiveVoteTally';
+import { cacheGet, cacheSet } from 'lib/cache';
 
 export default withApiHandler(async (req: NextApiRequest, res: NextApiResponse) => {
   const network = (req.query.network as string) || DEFAULT_NETWORK.network;
   invariant(isSupportedNetwork(network), `unsupported network ${network}`);
 
   const chief = getContracts(networkNameToChainId(network), undefined, undefined, true).chief;
+
+  const cacheKey = `executive-vote-tally-${chief.address}`;
+
+  const cached = await cacheGet(cacheKey, network);
+
+  res.setHeader('Cache-Control', 's-maxage=300, stale-while-revalidate');
+
+  if (cached) {
+    res.status(200).json(cached);
+    return;
+  }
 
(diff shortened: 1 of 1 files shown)


tyler17:
looks good, do we want to just merge to develop for now?

adamgoth (commented):
Think we should delete the cache after we detect an exec vote tx has been mined (similar to what we added for polls)?

adamgoth (changes requested):
I think you might have some changes you forgot to push. I see you moved cache into a module but I don't see the files for it. Also looks like there is a conflict to resolve.

rafinskipg:
> I think you might have some changes you forgot to push. I see you moved cache into a module but I don't see the files for it. Also looks like there is a conflict to resolve.

you are right, there's some problem detecting the modules/cache when i did git add, unfortunately i have the changes in the other computer, will get to it next week.

b-pmcg (dismissed):
LGTM, only question is whether doing invalidation in `pending` or `mined` is better?

pages/dashboard/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+          cacheKey
+        })
+      });
+      setLoading(false);
+      toast.success('Cache cleared');
+    } catch (e) {
+      setLoading(false);
+      toast.error('Error invalidating cache');
+    }
+  };
```
adamgoth, line 41: Should we export this and reuse it where I left previous comment?

modules/executive/components/VoteModal/DefaultView.tsx
```diff
@@ -157,6 +158,19 @@ export default function DefaultVoteModalView({
...
+          fetchJson(`/api/cache/invalidate?network=${network}`, {
+            method: 'POST',
+            headers: {
+              'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+              cacheKey: executiveSupportersCacheKey
+            })
+          });
+        }, 30000);
```
adamgoth, line 173: Could be useful to create helpers for these invalidation calls so that we can easily reuse and refactor

modules/polling/context/BallotContext.tsx
```diff
@@ -231,14 +232,21 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
...
+              method: 'POST',
+              headers: {
+                'Content-Type': 'application/json'
+              },
+              body: JSON.stringify({
+                cacheKey: getPollTallyCacheKey(parseInt(pollId))
+              })
+            });
+          }, 60000);
+        });
```
adamgoth, line 249: Here too

adamgoth (commented):
Pushed a few changes to remove unused imports and add a button to invalidate github execs. Let me know what you think about my other comments.

adamgoth (dismissed):
LGTM 💸

adamgoth (approved):

PR #541, opened 2022-06-28
add test to make sure we don't use bignumber.js directly

tyler17:


b-pmcg:
That's awesome, I didn't think about using glob. Nice catch on the test files as well. I was looking into the checkly errors and it appears to be a URL issue, doesn't seem like it was introduced here, but best to make sure.

b-pmcg (approved):

PR #543, opened 2022-06-29
Delegate banner updates

adamgoth:
### What does this PR do?

Updates the migration banner to check for both delegators who need to migrate and delegates who need to create new contracts


rafinskipg (approved):
LGTM but check the pages/migration/delegator there's an use of the previous hook

PR #544, opened 2022-06-29
Delegators expired

rafinskipg:
Adds logic in the delegators page to showcase delegates that are expiring soon.

The logic checks that the user had delegated to those delegates before.

Created a new component that changes in appearance if the delegate has expired or not, created with the calendar icon

image

Figma proto reference: https://www.figma.com/proto/MEkGk2IGAj4pQnFTPXHPFj/Governance-2022?node-id=2932%3A14650&scaling=min-zoom&page-id=2581%3A12305&starting-point-node-id=2610%3A12326


modules/delegation-migration/components/DelegateExpiryDate.tsx
```diff
@@ -0,0 +1,75 @@
...
+    },
+    redCalendar: {
+      fill: '#FF0000',
+      stroke: '#FF0000'
+    },
+    activeCalendar: {
+      fill: 'primary',
+      stroke: 'primary'
+    }
+  };
```
adamgoth, line 21: Do we not have variables for these colors? We should use whatever is closest instead of always adding slight variations.

pages/migration/delegator.tsx
```diff
@@ -39,6 +77,68 @@ export default function DelegateMigrationPage(): React.ReactElement {
...
+                <Text>There are no delegates that are about to expiry.</Text>
+              </Box>
+            )}
+          </Box>
+
+          <Box>
+            <Box>
+              <Text as="h1">Renewed contracts by your previous delegates</Text>
+              <Text as="p" variant="body">
+                Please delegate your MKR to the new contracts, one by one. Might require aproval transactions.
```
adamgoth, line 128: I know this is from the design but we know for a fact that approval transactions will be required, so we can update this copy. I think Deniz will review it all though.

PR #545, opened 2022-07-04
Migrate delegate page

adamgoth:
### What does this PR do?

Adds the UI for the steps for a delegate to migrate to a new contract


modules/migration/hooks/useMigrationStatus.tsx
```diff
@@ -2,12 +2,18 @@ import { useDelegatedTo } from 'modules/delegates/hooks/useDelegatedTo';
...
+export function useMigrationStatus(): {
   isDelegatedToExpiredContract: boolean;
   isDelegatedToExpiringContract: boolean;
   isDelegateContractExpired: boolean;
   isDelegateContractExpiring: boolean;
+  newOwnerAddress?: string;
+  newOwnerConnected: boolean;
+  previousOwnerAddress?: string;
+  previousOwnerConnected: boolean;
+  newOwnerHasDelegateContract: boolean;
```
rafinskipg, line 16: can we separate this into another hook for delegates? because this hook is used by delegators and delegates and it feels like a mix of things now. 

I think the first 4 variables have enough entity to be considered outside the migration status

adamgoth: We can separate them but I'm not sure how you are proposing to separate them? Which ones to include and which to separate?

PR #546, opened 2022-07-04
Link delegate addresses request

adamgoth:
### What does this PR do?

Adds signature and discord for request to link previous + new delegate owner addresses


pages/api/migration/link.ts
```diff
@@ -0,0 +1,31 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import withApiHandler from 'modules/app/api/withApiHandler';
+import { config } from 'lib/config';
+
+async function postRequestToDiscord(content: string) {
+  console.log({ content });
+  const resp = await fetch(config.MIGRATION_WEBHOOK_URL, {
```
rafinskipg, line 7: can we make that if the webhook is not configured returns a error message indicating that needs to add the env variable?

adamgoth: Good call. Added it.

PR #550, opened 2022-07-05
Delegator flow

rafinskipg:
Add empty states and use logic of previous/next delegate for delegators
image


adamgoth (commented):
Screen Shot 2022-07-05 at 2 28 47 PM

Not sure why but the circle shows purple for me. 

Also we should make sure the "Action Required" only shows when there is action required.

pages/migration/delegator.tsx
```diff
@@ -58,47 +64,40 @@ export default function DelegateMigrationPage(): React.ReactElement {
...
+              </Text>
+              <Text as="p" sx={{ color: 'onSecondary' }}>
+                Please migrate your MKR by undelegating from the expiring/expired contracts and redelegating
+                to the new contracts.
+              </Text>
+              <Text as="p" sx={{ color: 'onSecondary' }}>
+                Find in this page any expiring/expired delegate contracts you delegated MKR to, and requires
+                migration.
+              </Text>
+            </Box>
```
adamgoth, line 93: IMO this should be some sort of reusable Alert or something that we already have. Maybe we can check with Tiago.

adamgoth (approved):
Like you said, we'll want to do more thorough testing using mocked API responses once we have all the pieces together, but looks good to me for now 🚀

PR #551, opened 2022-07-06
Combine vote history, comments with previous contract

adamgoth:


pages/api/address/[address]/stats.ts
```diff
@@ -78,28 +79,43 @@ import { getVoteProxyAddresses } from 'modules/app/helpers/getVoteProxyAddresses
 export default withApiHandler(async (req: NextApiRequest, res: NextApiResponse<AddressAPIStats>) => {
   const network = (req.query.network as string) || DEFAULT_NETWORK.network;
   const tempAddress = req.query.address as string;
+  const delegate = req.query.delegate as string;
+  const prevDelegateAddress = req.query.prev as string;
```
Comment (rafinskipg) [inline pages/api/address/[address]/stats.ts:83]: I would suggest we change this endpoint to `/api/address/stats` and to receibe a JSON with an array of addresses as a POST method, or just queryparameters as a GET method

`/api/address/stats?address=0xasds&address=0x123&address=0x12321` this way the api endpoint is more flexible and admits N number of addresses

rafinskipg (changes requested):

rafinskipg (approved):

PR #552, opened 2022-07-07
delegate expired modal

rafinskipg:
https://www.figma.com/file/MEkGk2IGAj4pQnFTPXHPFj/Governance-2022?node-id=3134%3A15981

image


lib/theme/colors.js
```diff
@@ -34,8 +34,8 @@ export const colors = {
   notice: '#EB7720',
   noticeAlt: '#FFFBEF',
   onNotice: '#EB7720',
-  warning: '#CB532D',
-  warningAlt: '#F75524',
+  warningAlt: '#CB532D',
+  warning: '#F75524',
```
adamgoth, line 38: What's the reason for swapping these? We should make sure that other components aren't affected.

adamgoth (approved):
LGTM, but couple small questions.

Is this the correct placement for the expiration date?
Screen Shot 2022-07-11 at 9 07 18 AM

Screen Shot 2022-07-11 at 9 07 37 AM

lib/theme/colors.js
rafinskipg: i didn't see any other warningAlt being used

PR #553, opened 2022-07-08
Fetch landing pages delegates from clientside

rafinskipg:
Fetch landing page delegates from clientside, which will use API cache


b-pmcg (approved):
I approve this message

PR #555, opened 2022-07-11
Add is renewed logic to delegated to api call

rafinskipg:
- Adds "isRenewed" to delegated-to response
- use that logic to show banner


adamgoth (approved):
LGTM 💸

PR #557, opened 2022-07-11
add delegate reference in account detail

rafinskipg:
image


modules/address/components/AddressDetail.tsx
```diff
@@ -38,6 +39,32 @@ export function AddressDetail({ addressInfo }: { addressInfo: AddressApiResponse
...
+        <Box>
+          <Box sx={{ pl: [3, 4], pr: [3, 4], pt: [3, 4] }}>
+            <Text
+              as="p"
+              sx={{
+                fontSize: 4,
+                mb: 3,
+                fontWeight: 'semiBold'
+              }}
+            >
```
adamgoth, line 52: It would be nice to get away from using these types of custom Text components and use some type of heading variant or what not. This won't auto-scale with different screen sizes like our variants will. I'm guessing you just copied this from somewhere else in the component though 😅 In the future we should do some sort of audit.

PR #564, opened 2022-07-14
Add expiry to localstorage

rafinskipg:
Add a expiry key to localstorage, if a element has a expiry date set, when fetching it it will return null and delete it

### DIFF
--- a/modules/app/client/storage/localStorage.ts
+++ b/modules/app/client/storage/localStorage.ts
@@ -5,6 +5,17 @@ function hasLocalStorage(): boolean {
 export const localStorage = {
   get: (key: string, json?: boolean): any => {
     if (hasLocalStorage()) {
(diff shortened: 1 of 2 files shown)


modules/app/client/storage/localStorage.ts
```diff
@@ -25,9 +36,12 @@ export const localStorage = {
       window.localStorage.removeItem(key);
     }
   },
-  set: (key: string, value: string): void => {
+  set: (key: string, value: string, expiryMS?: number): void => {
     if (hasLocalStorage()) {
       window.localStorage.setItem(key, value);
+      if (expiryMS) {
+        window.localStorage.setIte(`${key}-expiry`, (Date.now() + expiryMS).toString());
```
adamgoth, line 43: should be `setItem` (typo)

PR #573, opened 2022-07-19
Cache in address detail, 10 minutes

rafinskipg:
Add cache to address detail, reducing requests for delegates and any other address

Add cache to get comment transaction to 5 minutes

Add cache to address stats page to 10 minutes

Add cache to address delegation history 10 minutes

Add cache to the delegates response to 10 minutes, not only to the github delegates


tyler17 (dismissed):

tyler17 (commented):
Should we also invalidate the executive supporters cache after a user votes in an executive?

PR #581, opened 2022-07-29
Feature/sc 1641/upgrade web3 react package to v8

adamgoth:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1641/upgrade-web3-react-package-to-v8

### What does this PR do?
Updates web3-react from v6 to v8
Adds gnosis safe as an option

### Steps for testing:
Try the different wallet options
Try switching and disconnecting etc


b-pmcg:
I noticed that when clicking "disconnect" from the Account box, and then refreshing the page the app will eagerly connect the account again. Would removing the connection from the array in `useOrderedConnections` be enough to prevent that?

next.config.js
```diff
@@ -15,15 +15,32 @@ const securityHeaders = [
...
 
   // adds x-frame-options
-  {
-    key: 'X-Frame-Options',
-    value: 'SAMEORIGIN'
-  },
+  // remove for safe app to load
+  // {
+  //   key: 'X-Frame-Options',
+  //   value: 'SAMEORIGIN'
```
rafinskipg, line 21: Should we restore this?

rafinskipg (dismissed):
waiting on the link to the modified package from safe-react

next.config.js
adamgoth: We won't be able to load the app in the safe app iframe with it set. 

> if you specify SAMEORIGIN, you can still use the page in a frame as long as the site including it in a frame is the same as the one serving the page.

I'm not sure when or why it got added. I guess click-jacking is perhaps the security concern.

https://www.keycdn.com/blog/x-frame-options

b-pmcg (approved):
LFG!

PR #585, opened 2022-08-03
Singleton provider api

b-pmcg:
### What does this PR do?
Exports a `contracts` SDK singleton based on the type of contracts needed. They might be:
1. a set of read only contracts using a `RpcBatchProvider`
2. a set of account-connected contracts using a signer provided by `web3-react`
3. either of the above keyed by a custom API key

So at most there would be 2 providers for each API key that we use to connect to a node.

I also split the `readOnly` and regular providers into separate files for readability & simplicity.

### What's left?
- Update our usage of `getEthersContracts`
  - we can move `voteProxy` and `voteDelegate` contracts to our SDK instead of using `getEthersContracts` by providing an initial address (to download the ABI) and then we must call `.connect()` on the contracts with any new/given address. ie, for each delegate or proxy address, we call `.connect()`. In the interest of time I didn't implement this bc our current system still works, and it feels like a separate initiative.

### Steps for testing:
- Use the app as normal, ideally we won't see a difference in performance, but (hopefully) we'll see a reduction in RPC calls.
- Make sure all `write` calls still work as expected, especially after switching accounts and network

### Why:
- because creating a provider each time (with `getDefaultProvider`
 or using `new`) allocates new memory, and in a loop where it happens many times, it could impact performance and send unnecessary calls with zombie threads that aren't garbage collected
- because if there is a re-render of a component after an initial call, a new provider is created and a new call is made. But if the provider is a singleton, it might not need to re-fetch the data
- there may be some initialization calls that happen each time a new provider is created
- because if there is any caching going on at the provider level, a new instance wouldn’t share that cache

### Research
- [This suggests](https://ethereum.stackexchange.com/questions/107565/best-practice-for-single-or-multiple-instances-of-provider-for-different-types-o) that using a singleton is better for performance
    - new memory is allocated each time a provider is created
    - `getDefaultProvider` indeed creates a new provider since the constructor is called
- if we get the provider/contract instance in the API, isn’t it creating a completely new instance with each api/page? [This](https://stackoverflow.com/questions/65092664/multiple-singleton-instances-across-files-in-typescript-and-nextjs) suggests in production, as long as you are importing from the same singleton file, then it is compiled as a single Serverside/api rather than each page on-demand (as it is with development)
    - tldr: import the provider inside each API page/file from the same location, and it should be a singleton
        - additional discussion re: prod vs development splitting [here](https://www.reddit.com/r/nextjs/comments/eiykfc/share_database_connection_in_nextjs_api_routes/) and [here](https://stackoverflow.com/questions/71495754/nextjs-pages-api-routes-are-compiled-at-runtime)


adamgoth (commented):
I'm having some issues sending transactions locally. On goerli, if I try to approve a delegate contract, I don't see anything pop up. I also don't see any console errors. 

When I attempted to submit a poll ballot, I immediately got a "transaction failed" toast before metamask even opened.

I haven't had enough time to dig in any further, but just wanted to give a heads up.

b-pmcg:
> I'm having some issues sending transactions locally. On goerli, if I try to approve a delegate contract, I don't see anything pop up. I also don't see any console errors.
> 
> When I attempted to submit a poll ballot, I immediately got a "transaction failed" toast before metamask even opened.
> 
> I haven't had enough time to dig in any further, but just wanted to give a heads up.

Thanks for testing, I figured out the issue. It wasn't triggering the SDK rebuild in the case where we need a signer and don't have one. This only affected cases when we used a new API key (eg. delegates), because the contracts set connected with that key weren't being rebuilt. Pushed up a fix (and some error handling improvements that I added while I was debugging).

adamgoth:
> > I'm having some issues sending transactions locally. On goerli, if I try to approve a delegate contract, I don't see anything pop up. I also don't see any console errors.
> > When I attempted to submit a poll ballot, I immediately got a "transaction failed" toast before metamask even opened.
> > I haven't had enough time to dig in any further, but just wanted to give a heads up.
> 
> Thanks for testing, I figured out the issue. It wasn't triggering the SDK rebuild in the case where we need a signer and don't have one. This only affected cases when we used a new API key (eg. delegates), because the contracts set connected with that key weren't being rebuilt. Pushed up a fix (and some error handling improvements that I added while I was debugging).

Awesome. Seems to be working well for me locally now. Seeing some unit tests fail in CI on static page regeneration timing out but going to merge into w3r stuff because I don't think it's related to your changes.

PR #593, opened 2022-08-09
Update README instructions for running locally

adamgoth:


.env.sample
```diff
@@ -1,9 +1,15 @@
...
+
+INFURA_KEY=
+ALCHEMY_KEY=
+ETHERSCAN_KEY=
+POCKET_KEY=
+GOERLI_FORK_API_KEY=
+GITHUB_TOKEN=
+MONGODB_URI=
+MONGODB_COMMENTS_DB=
 USE_CACHE=true
```
rafinskipg, line 11: can we add redis_url?

rafinskipg (dismissed):

.env.sample
adamgoth: Good catch, added

PR #600, opened 2022-08-24
executive page build improvements

b-pmcg:
### What does this PR do?
- Fetch all executives in a single call rather than loop
- Use eth-sdk for spell contracts
- Reduce limit of executives to fetch/prebuild


modules/executive/api/fetchExecutives.ts
```diff
@@ -21,29 +22,25 @@ export async function getGithubExecutives(network: SupportedNetworks): Promise<C
...
+  const githubRepo = {
+    owner: 'makerdao',
+    repo: 'community',
+    page: 'governance/votes'
+  };
 
-  const proposals = await Promise.all(
-    proposalUrls.map(async (proposalLink): Promise<CMSProposal | null> => {
+  const githubResponse = await fetchGithubGraphQL(githubRepo, allGithubExecutives);
+  const proposals = githubResponse.repository.object.entries
```
adamgoth, line 32: 🔥

PR #612, opened 2022-09-01
Gasless selection flow refactored

rafinskipg:
Adds the new flow to switch between gasless and standard signature

image


modules/polling/components/review/ReviewBox.tsx
```diff
@@ -1,40 +1,357 @@
...
+                  disabled={!ballotCount || !!(transaction && transaction?.status !== 'error')}
+                  sx={{ mt: 3, width: '100%' }}
+                >
+                  Proceed with Standard submission
+                </Button>
+                <Box>
+                  <Flex sx={{ alignItems: 'center', justifyContent: 'center', mt: 3 }}>
+                    <Text>You pay the gas fee.</Text>
+                  </Flex>
+                  <Box>
```
adamgoth, line 216: Maybe an unnecessary `<Box />` around this `<Flex />`?

adamgoth (commented):
LGTM but I noticed one bug. Not sure if it's from this PR or previous work.

If the network call to the comments API endpoint fails, the entire review box disappears.

Screen Shot 2022-09-02 at 9 16 33 AM

PR #614, opened 2022-09-02
Feature/sc 1723/create re usable function for presubmission

adamgoth:


modules/polling/helpers/hasMkrRequiredForGaslessVotingCheck.ts
```diff
@@ -0,0 +1,15 @@
+import { SupportedNetworks } from 'modules/web3/constants/networks';
+import { getMKRVotingWeight, MKRVotingWeightResponse } from 'modules/mkr/helpers/getMKRVotingWeight';
+import { MIN_MKR_REQUIRED_FOR_GASLESS_VOTING } from 'modules/polling/polling.constants';
+
+export async function hasMkrRequiredForGaslessVotingCheck(
+  voter: string,
```
rafinskipg, line 6: we can refactor this function to "hasMKRRequiredWeight" and pass the amount as a parameter, this way we can reuse to normal users also, passing "0.0" to check if they have more than 0 mkr

modules/polling/helpers/recentlyUsedGaslessVotingCheck.ts
```diff
@@ -0,0 +1,14 @@
...
+import { cacheGet } from 'modules/cache/cache';
+import { SupportedNetworks } from 'modules/web3/constants/networks';
+
+export async function recentlyUsedGaslessVotingCheck(
+  voter: string,
+  network: SupportedNetworks
+): Promise<boolean> {
+  const cacheKey = getRecentlyUsedGaslessVotingKey(voter);
+
+  const recentlyUsedGaslessVoting = await cacheGet(cacheKey, network);
```
rafinskipg, line 11: does this work on the client side ? :/ im confused

rafinskipg: if this works on the client side it means the cacheGet and env variable are exposed, it shouldn't be. 

this might need to be an API call to api/polling/vote?address=0x213 (GET call with some address parameters)

rafinskipg (changes requested):

modules/polling/helpers/recentlyUsedGaslessVotingCheck.ts
adamgoth: Haven't been able to verify cause gasless isn't working for me locally yet, but no, don't think it would work. `cacheGet` will return null since `config.USE_CACHE` wouldn't be accessible client-side. Will have to refactor to an endpoint.

rafinskipg (approved):

PR #620, opened 2022-09-07
Adam/0.9.0 qa

adamgoth:
Fixes some console errors

Adds consistent casing

### DIFF
--- a/modules/app/components/ClientRenderOnly.tsx
+++ b/modules/app/components/ClientRenderOnly.tsx
@@ -1,0 +1,19 @@
+import { useEffect, useState } from 'react';
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const ClientRenderOnly: React.FC<Props> = ({ children, ...delegated }) => {
+  const [hasMounted, setHasMounted] = useState(false);
+
+  useEffect(() => {
+    setHasMounted(true);
+  }, []);
+
+  if (!hasMounted) {
+    return null;
+  }
+
+  return <div {...delegated}>{children}</div>;
+};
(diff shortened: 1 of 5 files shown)


modules/app/components/ClientRenderOnly.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface Props {
+  children: React.ReactNode;
+}
+
+export const ClientRenderOnly: React.FC<Props> = ({ children, ...delegated }) => {
+  const [hasMounted, setHasMounted] = useState(false);
+
+  useEffect(() => {
+    setHasMounted(true);
+  }, []);
```
rafinskipg, line 12: i think here you can add a check "typeof window !== undefined" to make sure it's only client side

rafinskipg (approved):

modules/app/components/ClientRenderOnly.tsx
adamgoth: Wrapping `setHasMounted`?

modules/app/components/ClientRenderOnly.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const ClientRenderOnly: React.FC<Props> = ({ children, ...delegated }) => {
```
b-pmcg, line 7: nice idea. what is `delegated` in this case?

adamgoth: I guess any props you'd want to pass to the div. We don't need it here but don't think it hurts to include it.

modules/app/components/ClientRenderOnly.tsx
adamgoth: > i think here you can add a check "typeof window !== undefined" to make sure it's only client side

I think it's safe without it.  Did a bit of digging to confirm that `useEffect` only runs for the first time after mounting so will never get run on the server.

rafinskipg (approved):

PR #623, opened 2022-09-09
deprecate the custom apiKey in get contracts fixing wrong signer bug

b-pmcg:
### What does this PR do?
- Removes the custom alchemy api key for delegates and drastically simplifies `getContracts`.
- Fixes a bug where the signer was getting cached in an unintended way, causing an issue when switching accounts.

The custom apiKey was intended to help us separate and identify specific calls to delegate contracts in order to learn more about our Alchemy usage. Ultimately this method was kind of shoe-horned in, and only complicated things when we refactored `getContracts` into a singleton.

Probably an even better way to get granular data about our Alchemy usage and contract calls is to use a new Axiom library [next-axiom](https://github.com/axiomhq/next-axiom). Here we can add logging after the calls are made sending data to alchemy about the RPC call, and later do some data analytics.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,7 +1,6 @@
 type SystemConfig = {
   USE_CACHE: string;
   ALCHEMY_KEY: string;
-  ALCHEMY_KEY_DELEGATES: string;
   INFURA_KEY: string;
   ETHERSCAN_KEY: string;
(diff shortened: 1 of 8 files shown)


adamgoth (approved):
Looks good to me. Always nice to simplify things 👌

PR #625, opened 2022-09-12
Create key without date

adamgoth:


rafinskipg (changes requested):
this will cause an error because connectors and connections are not of the same type, 

mannuallyAddedConnectors do not have the "type" object key

PR #631, opened 2022-09-13
Validate and sanitize query params

rafinskipg:
Avoids injection of unwanted values on query parameters that are used for filtering 

Introduces a validateQueryParams method and some tests to ensure that it works correctly


adamgoth (approved):
Good stuff 🔥

Only suggestion is maybe we want put a logger.debug message when something gets invalidated and overwritten with a default value?

PR #634, opened 2022-09-13
refactor: Remove duplicated query for fetching votes for gasless, include arbiscan link on ui

rafinskipg:
Delete query that was doing the same thing,

deprecate optionId from votes

refactored code to use only one call to get votes and get rid of obsolete types

added links to arbiscan first implementation (not reviewed by tiago ) 
image


modules/polling/helpers/getPollTally.ts
```diff
@@ -2,33 +2,21 @@ import { SupportedNetworks } from 'modules/web3/constants/networks';
...
 
-  const endUnix = new Date(poll.endDate).getTime() / 1000;
-  const votesByAddress = await fetchVotesByAddressForPoll(poll.pollId, endUnix, network);
+  //const pollEnded = pollHasEnded(poll);
 
-  const tallyObject = {
-    ...tally,
-    votesByAddress
-  };
+  // cacheSet(cacheKey, JSON.stringify(tally), network, pollEnded ? ONE_DAY_IN_MS : THIRTY_SECONDS_IN_MS);
```
adamgoth, line 19: Did you forget to uncomment?

adamgoth (commented):
LGTM but once the build is fixed, we should compare API responses for tallies for some polls

modules/polling/helpers/getPollTally.ts
rafinskipg: no, i want to continue working on it

PR #635, opened 2022-09-14
Add timestamps to ballot votes and tx hash

rafinskipg:
Adds timestamps to ballot objects, onload if the timestamp is greater than 1 day, remove it.

When a tx is pending, store the txhash on the ballots, onload if they have a txhash, remove them


modules/polling/context/BallotContext.tsx
```diff
@@ -109,26 +129,25 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
...
-  const addVoteToBallot = (pollId: number, ballotVote: BallotVote) => {
+  const addVoteToBallot = (pollId: number, ballotVote: Partial<BallotVote>) => {
     setTxId(null);
     setCommentSignature('');
     const newBallot = {
       ...ballot,
-      [pollId]: ballotVote
+      [pollId]: {
+        ...ballotVote,
+        timestamp: Date.now()
```
adamgoth, line 139: I don't think you're using timestamp for anything, can remove it here and in `updateVoteFromBallot` and from the type def

adamgoth (commented):
Looks good and works for me locally, just need to remove timestamp I think

adamgoth (approved):

PR #639, opened 2022-09-15
Gasless eligibility criteria

adamgoth:
Adds checks and UI for gasless submission eligibility 

Screen Shot 2022-09-15 at 12 32 47 PM

Screen Shot 2022-09-15 at 12 32 51 PM


lib/theme/icons.js
```diff
@@ -1159,5 +1159,16 @@ export const icons = {
         />
       </g>
     )
+  },
+  close: {
```
rafinskipg, line 1163: image
I think we already have this icon,

rafinskipg (changes requested):

lib/theme/icons.js
adamgoth: Just checked. In the screenshot above, it's a `Close` component imported from theme-ui, and it's a button element. So I think we need the icon (non-button).

rafinskipg: but is in theme-ui it means the icon has to be there somewhere right? 
image

rafinskipg: Anyway, if it changes a lot of our coding style and the way to theme the component or import the icon, i'm in favor of adding the new icon instead of importing the old from theme-ui

PR #642, opened 2022-09-16
Refactoring api for comments on gasless

rafinskipg:
Changes to get and add comments in gasless , it puts a property on the comment object indicating that is gasless, but it mantains the origin network.


tyler17:
Agreed this is a better way to do it

PR #644, opened 2022-09-16
backoffRetry when fetching gasless transaction in case it hasn't been…

b-pmcg:
… picked up by network yet

### What does this PR do?
Due to a possible delay with the relayer, sometimes the tx doesn't appear on the network before we try to get it with the provider. This results in the tx object being null & an error. This PR uses backoffRetry to attempt to fetch the tx again if it is not found at first.


tyler17:
😄 👍

tyler17 (approved):

rafinskipg (approved):

PR #645, opened 2022-09-19
Reuse voting account trhough app and fix precheck gasless

rafinskipg:
We had duplicated code that checked which was the right account to use for voting, i removed it and added to the "useACcount" hook, so we can just access the right voting account.

The "precheck" for gasless was using the wrong address, and failed if you used a delegate contract. Fixed now.

Simplified the "useLockedMKR" hook to only have 1 parameter


tyler17:
Nice catch, looks good to me.

tyler17 (approved):

PR #646, opened 2022-09-19
Gasless comments working

rafinskipg:
- Add goerli hardcoded relayer address (remember to remove or update this check before going to production)
- refactor some functions to reuse the "getDelegateContractAddress" method
- make sure that the vote api checks against the correct address (delegate or proxy ) to know if the user voted before on a poll

### DIFF
--- a/modules/comments/types/comments.ts
+++ b/modules/comments/types/comments.ts
@@ -35,7 +35,10 @@ export type CommentsAPIResponseItem = {
 };
 
 export type ParsedExecutiveComments = {
-  comment: Omit<ExecutiveComment, 'voterWeight'> & { voterWeight: BigNumber };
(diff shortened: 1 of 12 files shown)


modules/comments/types/comments.ts
```diff
@@ -35,7 +35,10 @@ export type CommentsAPIResponseItem = {
 };
 
 export type ParsedExecutiveComments = {
-  comment: Omit<ExecutiveComment, 'voterWeight'> & { voterWeight: BigNumber };
+  comment: Omit<ExecutiveComment, 'voterWeight'> & {
+    voterWeight: BigNumber;
+    gaslessNetwork?: SupportedNetworks;
```
rafinskipg, line 40: needed to add this property here to mantain the same type with the pollComments so the comment component does not complain, we will think about another approach in the future

PR #653, opened 2022-09-22
handle errors from vote API as well as RPC errors from wallet

b-pmcg:
### Link to Shortcut ticket:
[sc-1773]
[sc-1776]

### What does this PR do?
Handle errors that come from our API and parse them for the error message to display. Also handle RPC and provider related errors with [eth-rpc-errors](https://github.com/MetaMask/eth-rpc-errors) lib

Also disable the comments textarea after submitting the tx to the relayer, but before we get the txId.


tyler17 (approved):
Looks like some tests may need to be updated, but otherwise looks good

rafinskipg (changes requested):
The appi should return a error object,
```
{
  error: 'Something'
}
```

At least that is how is handled on the rest of apy endpoints

rafinskipg (approved):

adamgoth (approved):

PR #656, opened 2022-09-26
Change fetchJSON implementation to return the parsed error, assume ou…

rafinskipg:
…r API always return a error object

Our API has always to return a JSON object with an error key, 
- Change fetchJSON implementation to reject with an object instead of a string
- fetchJSON checks if the api returns the error key and parses it
- change transaction store to properly extract the error codes from metamask or provider
- handle both error separately


pages/api/polling/vote.ts
```diff
@@ -45,68 +64,68 @@ export default withApiHandler(
       const { voter, pollIds, optionIds, nonce, expiry, signature, network, secret } = req.body;
 
       if (typeof voter !== 'string' || !voter) {
-        const error = { message: { error: API_VOTE_ERRORS.VOTER_MUST_BE_STRING } };
-        postError(JSON.stringify({ ...error.message, ...req.body }));
+        const error = { error: API_VOTE_ERRORS.VOTER_MUST_BE_STRING };
```
b-pmcg, line 67: I think the error object typically has a message property instead of `error`. Most examples I've seen look more like this:
```
error: {
    code: 1150,
    message: 'bla bla bla'
  }
}
```

lib/fetchJson.ts
```diff
@@ -2,6 +2,13 @@ export async function fetchJson(url: RequestInfo, init?: RequestInit): Promise<a
   const response = await fetch(url, init);
   const json = await response.json();
 
-  if (!response.ok) throw new Error(`${response.statusText}: ${json.error?.message || JSON.stringify(json)}`);
+  if (!response.ok) {
+    return Promise.reject({
```
b-pmcg, line 6: Nice, this was driving me crazy. This is a better way to handle it, just have to make sure our implementations will handle this schema.

modules/web3/helpers/errors.ts
```diff
@@ -1,9 +1,27 @@
...
   }
-  return error;
+
+  if (message.includes('user rejected transaction')) {
+    return USER_REJECTED;
+  }
+
+  // First check if it's a Metamask error
+  if (error['code']) {
+    const extracted = getMessageFromCode(error['code']);
```
b-pmcg, line 20: I noticed `getMessageFromCode` will return a fallback message if it can't find the matching code, something like `Unspecified error message. This is a bug, please report it.`. We don't want to display that so that's why I check if the code is in the list of provider/rpc codes.

modules/web3/helpers/errors.ts
```diff
@@ -1,9 +1,27 @@
...
+export function parseTxError(error: Error): string {
+  const defaultError = 'Something went wrong with your transaction.';
+
+  const message = error.message ? error.message : defaultError;
+  if (message.includes("sender doesn't have enough funds to send tx")) {
     return TX_NOT_ENOUGH_FUNDS;
   }
-  return error;
+
+  if (message.includes('user rejected transaction')) {
```
b-pmcg, line 14: also why not utilize the error code here? In case the message changes someday. We can still return a custom message if that's the goal. `if (error.code === 4001) myCustomMessage`

PR #657, opened 2022-09-26
add api function for fetching relayer balance

b-pmcg:
TODO: hide text or show spinner when loading


pages/api/polling/relayer-balance.ts
```diff
@@ -0,0 +1,32 @@
...
+import { DEFAULT_NETWORK } from 'modules/web3/constants/networks';
+import invariant from 'tiny-invariant';
+
+export default withApiHandler(
+  async (req: NextApiRequest, res: NextApiResponse) => {
+    try {
+      const network = (req.query.network as string) || DEFAULT_NETWORK.network;
+      invariant(isSupportedNetwork(network), `unsupported network ${network}`);
+
+      const balance = await getRelayerBalance(network);
```
b-pmcg, line 16: we could just do a regular `provider.getBalance` in the frontend with the hardcoded address too. Maybe that would be easier but it's rather easy to get the address from the relayer signer too. 🤷

modules/polling/helpers/getRelayerBalance.ts
```diff
@@ -0,0 +1,28 @@
+import { getArbitrumGoerliTestnetSdk, getArbitrumOneSdk } from '@dethcrypto/eth-sdk-client';
+import { config } from 'lib/config';
+import { ArbitrumSdkGenerators } from 'modules/web3/types/contracts';
+import { SupportedNetworks } from 'modules/web3/constants/networks';
+import { DefenderRelayProvider, DefenderRelaySigner } from 'defender-relay-client/lib/ethers';
+import { formatEther } from 'ethers/lib/utils';
+
+export const relayerCredentials = {
+  mainnet: { apiKey: config.DEFENDER_API_KEY_MAINNET, apiSecret: config.DEFENDER_API_SECRET_MAINNET },
+  goerli: { apiKey: config.DEFENDER_API_KEY_TESTNET, apiSecret: config.DEFENDER_API_SECRET_TESTNET }
```
rafinskipg, line 10: maybe we should reuse the same relayer credentials from `getArbitrumPollingContractRelayProvider`

pages/api/polling/relayer-balance.ts
rafinskipg: is really nice what you did with the relayer api, is more reliable

rafinskipg (changes requested):

pages/api/polling/relayer-balance.ts
tyler17: Yeah this sounds good too!

tyler17 (approved):

modules/polling/helpers/getRelayerBalance.ts
b-pmcg: 💯

rafinskipg (approved):

pages/api/polling/relayer-balance.ts
```diff
@@ -0,0 +1,32 @@
...
+        [...Object.values(ERROR_CODES.provider), ...Object.values(ERROR_CODES.rpc)].includes(err['code'])
+          ? getMessageFromCode(err['code'])
+          : err.message;
+      logger.error(errorMessage);
+      return res.status(400).json({
+        error: errorMessage
+      });
+    }
+  },
+  { allowPost: true }
```
adamgoth, line 31: I think this can be removed since this doesn't look like it's an endpoint for POST requests

pages/api/polling/precheck.ts
```diff
@@ -29,15 +30,17 @@ export default withApiHandler(async (req: NextApiRequest, res: NextApiResponse)
   const cacheKey = getRecentlyUsedGaslessVotingKey(voter);
 
   // TODO add a check to see if user has already voted in polls?
```
adamgoth, line 32: Not part of your updates but noticed this is a comment that can be removed now (if you make any additional PR updates)

PR #658, opened 2022-09-27
Fixes for QA and move all links to etherscanlink component

rafinskipg:
Fixes on styles, refactor all links to etherscan to reuse component so we can handle logic to display gasless info
image
image
image
image


modules/address/components/AddressDelegatedTo.tsx
```diff
@@ -16,6 +14,7 @@ import AddressIconBox from './AddressIconBox';
 import { parseUnits } from 'ethers/lib/utils';
 import { formatValue } from 'lib/string';
 import { DateWithHover } from 'modules/app/components/DateWithHover';
+import EtherScanLink from 'modules/web3/components/EtherScanLink';
```
adamgoth, line 17: Not a big deal but Etherscan doesn't use camelcasing
Screen Shot 2022-09-28 at 9 15 04 AM

adamgoth (approved):

PR #660, opened 2022-09-28
stats relayer

rafinskipg:
image


adamgoth:
Maybe it makes sense to add this to the /dashboard instead.

That way it's password-gated and it also could be relevant for gov alpha to see balances.

PR #661, opened 2022-09-29
Exec cache

adamgoth:
### What does this PR do?

Prior to this, we were only clearing the cache key for github executives.

This did not clear the proposals cache key, so exec lists stayed the same.


rafinskipg (approved):
Lgtm

PR #664, opened 2022-09-30
Add new class API error and change API to return the same error struc…

rafinskipg:
- [x] Added new  ApiError class to handle api errors 
- [x] withApiHandler now handles the errors returned by this internal class and formats the response, also logs all the errors with good format
- [x] fetchJSON handles errors with this format and gets the right message
- [x] GQL requests uses ApiError to return formatted error info
- [x] Custom error500 page to show more info about the error
- [ ] Do not break app if GQL requests fails

image

### DIFF
--- a/modules/app/components/ErrorPage.tsx
+++ b/modules/app/components/ErrorPage.tsx
@@ -1,0 +1,21 @@
+import { Box, Text } from 'theme-ui';
+
+export default function ErrorPage({
+  statusCode,
+  title = 'An unexpected error has ocurred',
+  children
+}: {
+  statusCode: number;
+  title?: string;
+  children?: React.ReactElement;
+}): React.ReactElement {
+  return (
+    <Box>
+      <Text>{statusCode} </Text> | {title}
+      <Box>{children}</Box>
+      <Box>
+        <Text>For more information or help contact the Development and UX core unit at : Discord</Text>
+      </Box>
+    </Box>
+  );
+}
--- a/modules/app/components/ErrorPage.tsx
+++ b/modules/app/components/ErrorPage.tsx
@@ -1,4 +1,6 @@
-import { Box, Text } from 'theme-ui';
+import Head from 'next/head';
+import { Box, Flex, Text } from 'theme-ui';
+import { ExternalLink } from './ExternalLink';
 
 export default function ErrorPage({
   statusCode,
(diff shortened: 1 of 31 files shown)


modules/app/components/ErrorPage.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Box sx={{ maxWidth: '500px', margin: '0 auto', pt: '200px', paddingBottom: '200px' }}>
+        <Flex sx={{ justifyContent: 'center', alignItems: 'center' }}>
+          <Text sx={{ fontSize: 5, fontWeight: 'bold', mr: 2 }}>{statusCode}</Text> |{' '}
+          <Text sx={{ ml: 2 }}>{title}</Text>
+        </Flex>
+        <Box pt={3}>{children}</Box>
+        <Box pt={3}>
+          <Text sx={{ textAlign: 'center' }}>
+            For more information or help contact the Development and UX core unit at{' '}
+            <ExternalLink href="https://discord.gg/GHcFMdKden" title="Discord">
```
adamgoth, line 30: `For more information or help, please contact the Development & UX core unit on Discord` reads a little better.

modules/mkr/helpers/getMKRVotingWeight.ts
```diff
@@ -70,7 +70,7 @@ export async function getMKRVotingWeight(
   // otherwise, not proxy or delegate, get connected wallet balances
   const walletBalanceHot = await contracts.mkr.balanceOf(address);
   const chiefBalanceHot = await contracts.chief.deposits(address);
-
+  console.log('eeeeee', chiefBalanceHot.toNumber());
```
adamgoth, line 73: ✂️

pages/address/[address]/index.tsx
```diff
@@ -110,8 +110,18 @@ export default function AddressPage(): JSX.Element {
...
+  if (error && error.status === 404) {
+    return (
+      <PrimaryLayout>
+        <ErrorPage statusCode={404} title="This address does not exist" />
+      </PrimaryLayout>
+    );
+  } else if (error) {
+    return (
+      <PrimaryLayout>
+        <ErrorPage statusCode={500} title="Error fetching address information" />
```
Comment (adamgoth) [inline pages/address/[address]/index.tsx:122]: Are we able to pass `error.status` instead of hardcoding 500?

pages/api/address/stats.ts
```diff
@@ -84,9 +85,7 @@ export default withApiHandler(async (req: NextApiRequest, res: NextApiResponse)
       : (req.query.address as string[]);
 
   if (!req.query.address) {
-    return res.status(400).json({
-      error: 'Missing address'
-    });
+    throw new ApiError('Address stats, missing address', 401, 'Missing address');
```
adamgoth, line 88: 401 is typically used for `Unauthorized` where 400 is for `bad request`. Wouldn't 400 make more sense here?

adamgoth (commented):
Looks good. Just left a couple comments about small things.

pages/api/address/stats.ts
rafinskipg: yes! typo

PR #672, opened 2022-10-20
Test/cypress docker

b-pmcg:
### What does this PR do?

Adds support for a dockerized container of the gov polling db so we can run tests against a working database that's synced to the goerli chain. Helpful for for testing gasless voting. (See included [README](https://github.com/makerdao/governance-portal-v2/blob/test/cypress-docker/README.md#goerli-fork) for instructions). In particular, try the `review.spec.ts` test file as not all files are updated to work yet.

Still need to:

- [x] Get mock relayer working
- [x] update the code to use the `LOCAL_SPOCK_URL` graphql endpoint when running tests.
- [x] update review.spec.ts for the gasless UI changes
- [ ] ???
- [ ] profit


scripts/setup.js
```diff
@@ -22,59 +26,67 @@ async function main() {
...
-      to: keyPairs.addresses[i],
-      value: ethers.utils.parseEther('0.5')
-    });
-  }
+  // for (let i = 0; i < 50; i++) {
+  await ethSender.sendTransaction({
+    to: '0x8028Ef7ADA45AA7fa31EdaE7d6C30BfA5fb3cf0B',
+    value: ethers.utils.parseEther('5')
+  });
+  // }
```
adamgoth, line 34: Looks like this updates the script to only fund one wallet.

If that's the case, should we remove the old comment and commented out lines?

scripts/setup.js
```diff
@@ -22,59 +26,67 @@ async function main() {
...
   const mkrToken = new ethers.Contract(mkrAddress, ERC20_ABI, signer);
 
-  // Send 0.01 MKR to all addresses
-  for (let i = 0; i < 50; i++) {
-    await mkrToken.transfer(keyPairs.addresses[i], ethers.utils.parseEther('0.01'));
-  }
+  // // Send 0.01 MKR to all addresses
+  // for (let i = 0; i < 50; i++) {
+  //   await mkrToken.transfer(keyPairs.addresses[i], ethers.utils.parseEther('0.01'));
+  // }
```
adamgoth, line 50: Do we want this anymore?

adamgoth (approved):
Two small questions about the setup script. 

Great work! 🙌

scripts/setup.js
b-pmcg: Good call, I commented these out early on because I was having some issues with accounts being funded. I'll double check if they can be re-enabled in the interest of keeping the file as close to the original as possible (same as comment below)

b-pmcg: I reverted this back to what it was and tested that its working properly. I updated some of the logs to be more specific.

PR #675, opened 2022-10-26
Fixed ballot

rafinskipg:
IDEA: Fixed ballot follows you when scrolling
image
image


modules/polling/components/FixedBottomBallot.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <InternalLink href="/polling/review" title="Review & submit your ballot">
+            <Button
+              onClick={() => {
+                trackButtonClick('reviewAndSubmitBallot');
+              }}
+              variant="primary"
+              disabled={!ballotCount}
+              sx={{ cursor: !ballotCount ? 'not-allowed' : 'pointer', mr: 2 }}
+            >
+              Submit Your Ballot
```
adamgoth, line 42: Minor thing, but besides whatever feedback Tiago has, I think this is better to say "Review your ballot" since it takes you to the review page first before submitting.

PR #684, opened 2022-11-07
migrate to next13

rafinskipg:
Upgraded to nextjs and did some manual testing through the pages, seems to work ok. 

No added new features from nextjs


modules/app/stores/uiFilters.ts
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 
 */
 
-import create from 'zustand';
+import { create } from 'zustand';
```
adamgoth, line 9: Just wondering why this was necessary. Doesn't seem zustand version was updated.

adamgoth (approved):
We should do team-wide QA on develop before we go to prod because it's so many changes, but did a quick walkthrough the deployed URL and things looked good. Site seems quicker 💨

modules/app/stores/uiFilters.ts
rafinskipg: When running the portal it kept complaining that is deprecated and that we should import it like this now

torhector2 (approved):
💯

PR #686, opened 2022-11-07
allow for skipping discord error message in vote api

tyler17:
If you include skipDiscord in the request, discord won't be notified of the error.  This way TechOps can run a test that makes sure error messages are properly returned from the vote api without spamming our discord.

### Steps for testing:
Screen Shot 2022-11-07 at 5 15 03 PM


adamgoth (approved):
Good call 👍

PR #687, opened 2022-11-08
Fix/gasless

tyler17:
-Start racking gasless tranasctions before they gets mined
-don't check from address when verifying comment transactions, since gasless votes come from the relayer, not the voter
-fix comment filtering when fetching comments by address (pass in the right provider, and filter out invalid txs)

### DIFF
--- a/modules/polling/context/BallotContext.tsx
+++ b/modules/polling/context/BallotContext.tsx
@@ -419,9 +419,7 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
             () =>
               fetchJson(url).then(tx => {
                 // gaslessTx.status can be: 'pending' | 'sent' | 'submitted' | 'inmempool' | 'mined' | 'confirmed' | 'failed'
-                if (tx.status === 'mined') {
-                  // check if mined, if so, do nothing ????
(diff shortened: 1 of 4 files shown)


b-pmcg (approved):

modules/polling/context/BallotContext.tsx
```diff
@@ -409,18 +409,17 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
 
         const gaslessProvider = getGaslessProvider(network);
 
+        const voteTxCreator = () => getGaslessTransaction(gaslessProvider, gaslessTx.hash);
+        trackPollVote(voteTxCreator, getGaslessNetwork(network));
+
```
adamgoth, line 414: I had moved this to wait for `mined` status previously because the `getGaslessTransction`function was returning an error when the tx was just sitting the queue because the provider can't find it. If you click the "View on Arbiscan" right away and the tx is in the queue, you'll see that the tx isn't found. But I guess 99% of the time it will be relayed shortly after so probably not an issue.

tyler17: Ah interesting.  Cool that makes sense to me

PR #691, opened 2022-11-10
Signtypeddataethers

rafinskipg:
Remove the dependency of metamask package
- sign type data with ethers
- recover typed data with ethers
- remove unused type


adamgoth (approved):
Tested with ledger and MM, no issues

PR #700, opened 2022-11-15
add query variables for allCurrentVotes pagination

b-pmcg:

### DIFF
--- a/modules/gql/queries/allCurrentVotes.ts
+++ b/modules/gql/queries/allCurrentVotes.ts
@@ -1,15 +1,34 @@
 import { gql } from 'graphql-request';
 
 export const allCurrentVotes = gql`
-  query allCurrentVotes($argAddress: String!) {
-    allCurrentVotes(argAddress: $argAddress) {
-      nodes {
-        pollId
-        optionIdRaw
-        blockTimestamp
-        chainId
-        mkrSupport
-        hash
+  query allCurrentVotes(
+    $first: Int
+    $before: Cursor
+    $last: Int
+    $after: Cursor
+    $offset: Int
+    $filter: AllCurrentVotesRecordFilter
+    $argAddress: String!
+  ) {
+    allCurrentVotes(
+      first: $first
+      before: $before
+      last: $last
+      after: $after
+      offset: $offset
+      filter: $filter
+      argAddress: $argAddress
+    ) {
+      edges {
+        node {
+          pollId
+          optionIdRaw
+          blockTimestamp
+          chainId
+          mkrSupport
+          hash
+        }
+        cursor
       }
     }
   }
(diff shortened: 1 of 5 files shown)


modules/gql/queries/allCurrentVotes.ts
```diff
@@ -1,15 +1,34 @@
 import { gql } from 'graphql-request';
 
 export const allCurrentVotes = gql`
```
b-pmcg, line 3: The query in the gov polling db needed to be updated to accept the query variables such as `first`, and this query matches it. See the PR here https://github.com/makerdao/gov-polling-db/pull/127/files

modules/polling/api/fetchAllCurrentVotes.ts
```diff
@@ -1,27 +1,30 @@
...
   address: string,
-  network: SupportedNetworks
+  network: SupportedNetworks,
+  variables: AllCurrentVotesQueryVariables
 ): Promise<PollTallyVote[]> {
   const data = await gqlRequest({
     chainId: networkNameToChainId(network),
     query: allCurrentVotes,
-    variables: { argAddress: address.toLowerCase() }
+    variables
```
b-pmcg, line 17: Now that the query signature has been updated we can pass any of the variables we want to the query

modules/polling/api/fetchAddressPollVoteHistory.ts
```diff
@@ -10,7 +10,11 @@ export async function fetchAddressPollVoteHistory(
 ): Promise<PollVoteHistory[]> {
   // TODO: This is an innefective way to cross fetch titles and options. We should improve Spock DB to return the titles in the poll votes
   const pollsData = await getPolls({}, network);
-  const voteHistory = await fetchAllCurrentVotes(address, network);
+
+  // TODO: remove this mock pagination
+  const [first, offset] = [2, 0];
+  const queryVariables = { argAddress: address.toLowerCase(), first, offset };
+  const voteHistory = await fetchAllCurrentVotes(address, network, queryVariables);
```
b-pmcg, line 17: This is just an example of how pagination can work, incrementing the offset will give us the next group of votes. Maybe it's better to use `cursor` instead, that method is also available now.

modules/polling/hooks/useAllUserVotes.ts
```diff
@@ -12,10 +12,14 @@ type AllUserVotesResponse = {
 
 export const useAllUserVotes = (address?: string): AllUserVotesResponse => {
   const { network } = useWeb3();
+
+  // TODO: remove this mock pagination
+  const [first, offset] = [2, 0];
```
b-pmcg, line 17: Careful not to merge this with the mocked hardcoded values

b-pmcg:
> @b-pmcg are we able to close this PR? I assume @hernandoagf 's work has accomplished what you intended here but I'll let you confirm.

Yep, you're right, I'll close this one.

PR #713, opened 2023-01-19
Fix/UI elems, gas price indicator, mkr weight for delegate owners and change tooltips

hernandoagf:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1585/replace-sun-icon-with-moon-for-darkmode-toggle-in-menu
https://app.shortcut.com/dux-makerdao/story/1855/fix-gas-price-indicator
https://app.shortcut.com/dux-makerdao/story/1787/fix-mkr-weight-display-of-addresses-that-own-delegate-contracts
https://app.shortcut.com/dux-makerdao/story/1783/remove-your-vote-subheader-on-executive-detail-page
https://app.shortcut.com/dux-makerdao/story/1848/fix-overlap-of-embedded-media-on-poll-detail-page
https://app.shortcut.com/dux-makerdao/story/1836/deprecate-reach-ui-by-replacing-long-tooltips-with-dialogs

And also, added a skeleton component for some delegate balances. I think there's no ticket for it yet but it was discussed on Discord a couple of days ago: https://discord.com/channels/836646301986848768/882969853391163392/1064494148192833546

### Steps for testing:
1. On the header menu, switch between dark and light mode and make sure the icon updates accordingly.
2. On any of the pages, check the gas price indicator is showing in the header bar and that the gas price estimation is correct.
3. Delegate owner addresses used to show some MKR balance corresponding to the MKR in their delegate contract, e.g. https://vote.makerdao.com/address/0xa3f0abb4ba74512b5a736c5759446e9b50fda170#account-details, check that this is not the case anymore.
4. Verify that the `Your Vote` header is not present anymore on the voting section of any executive.
5. There's a couple of old polls that have large images on the poll details, e.g. https://vote.makerdao.com/polling/QmPEM83s#poll-detail, their images should adjust to the container size now.
6. Go to the delegates section and check that both `MKR delegated by you` and `Total MKR delegated` display skeleton components when loading instead of 0.
7. Tooltips changed for modals:
7.1. Polling contracts v1 and v2
7.2. Delegate participation and communication stats (on delegate cards of delegates section and on delegate metrics section of a delegate page)
7.3. Poll types (on each poll card of the polling section)
7.4. Finally, changed a couple of tooltips for them to be multi-line instead of one long line.


modules/app/components/SystemStatsSidebar.tsx
```diff
@@ -38,6 +39,42 @@ type StatField =
...
+              <Flex
+                sx={{
+                  flexDirection: 'column',
+                  justifyContent: 'center',
+                  alignItems: 'center'
+                }}
+              >
+                <Heading sx={{ mb: 3 }}>Polling contract versions</Heading>
+                <Text sx={{ textAlign: 'center' }}>
+                  - The latest version of the polling contract was deployed to enable batch voting, so users
```
adamgoth, line 63: Should we put something like V1: <description> and V2: <description> so we know which description corresponds to both? Since they both open the same modal?

modules/delegates/components/DelegateOverviewCard.tsx
```diff
@@ -29,12 +25,55 @@ import { CoreUnitModal } from './modals/CoreUnitModal';
...
+                  - Participation: The percentage of votes the delegate has participated in. It combines stats
+                  for polls and executives.
+                  <br />
+                  <br />- Communication: The percentage of votes for which the delegate has publicly
+                  communicated their reasoning in addition to voting. It combines stats for polls and
+                  executives.
+                  <br />
+                  <br />
+                  Both stats are updated weekly by the GovAlpha Core Unit.
+                </Text>
```
adamgoth, line 67: This is a minor thing, but from a semantic HTML perspective, I think each of these sections (participation, communication, and the last line) should be in their own `<Text as="p>{text}</Text>` components. Right now the whole thing shows as one span, but a span is usually used for inline elements to mark up a part of a text. Not for entire paragraphs.

Screenshot 2023-01-23 at 10 20 59 AM

modules/app/components/SystemStatsSidebar.tsx
adamgoth: See my comment below on the other modal text regarding `p` vs `span` for this content.

modules/polling/components/VotingWeight.tsx
```diff
@@ -73,9 +73,10 @@ export default function VotingWeight(): JSX.Element {
 
   const tooltipLabel = (
     <Box>
-      <Text as="p" sx={{ whiteSpace: 'normal' }}>
+      <Text as="p" sx={{ whiteSpace: 'normal', maxWidth: '400px' }}>
         {votingWeightCopy}
       </Text>
+      <br />
```
adamgoth, line 79: We can add `mb: 3` (short for marginBottom) or however much space we need below this, to the `sx` prop on the <Text />` element, rather than a `<br /`> element below it. I think it's better to use the style system where we can.

pages/polling/[poll-hash].tsx
```diff
@@ -53,7 +53,9 @@ import BoxWithClose from 'modules/app/components/BoxWithClose';
 
 const editMarkdown = content => {
   // hide the duplicate proposal title
-  return content.replace(/^<h1>.*<\/h1>|^<h2>.*<\/h2>/, '');
+  return content
+    .replace(/^<h1>.*<\/h1>|^<h2>.*<\/h2>/, '')
+    .replace(/(<img)(.*src=".*")(>)/g, '$1 width="100%"$2$3');
```
Comment (adamgoth) [inline pages/polling/[poll-hash].tsx:58]: Maybe add a comment above this new `replace` line so that people in the future know what it's for. Otherwise it seems a bit random and hard to guess why it's necessary 😅

adamgoth (commented):
I left a couple of small comments where I think small things could be improved, but overall it looks great!

adamgoth (approved):

PR #717, opened 2023-01-20
Endpoint validations

adamgoth:
Adds validations on endpoints parameters


pages/api/address/[address]/delegated-to.ts
```diff
@@ -6,28 +6,50 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
 export default withApiHandler(
   async (req: NextApiRequest, res: NextApiResponse<MKRDelegatedToAPIResponse>) => {
-    const network = (req.query.network as string) || DEFAULT_NETWORK.network;
+    // validate network
+    const network = validateQueryParam(
+      (req.query.network as SupportedNetworks) || DEFAULT_NETWORK.network,
+      'string',
+      {
+        defaultValue: null,
+        validValues: [SupportedNetworks.GOERLI, SupportedNetworks.GOERLIFORK, SupportedNetworks.MAINNET]
```
Comment (rafinskipg) [inline pages/api/address/[address]/delegated-to.ts:35]: i'm thinkingwe could add a "required" flag on the validateQueryParams function to make sure the value exists, and we would get some lines of code through the app

Comment (adamgoth) [reply pages/api/address/[address]/delegated-to.ts:35]: I was thinking about this too. Maybe we have another parameter for the error to throw too and the function can throw the if some validation fails. What do you think?

rafinskipg (approved):

PR #719, opened 2023-01-23
refactor validation function

rafinskipg:
Creates a function called validateAddress that resolves if an address is valid or fetches ENS resolution.

Removes files under pages/api that should be on helpers or similar.


adamgoth (approved):
Good stuff

PR #723, opened 2023-01-24
Fix mkr balances not showing for delegates

hernandoagf:
### What does this PR do?
It fixes the bug introduced during the last update that would not let delegates vote due to the MKR balance of their delegate contract not being fetched by passing a boolean value dynamically for the components that should display the MKR value of the owned delegate contract.

### Steps for testing:
If you have a delegate contract on Goerli, try to vote on polls.
The MKR balance of the delegate contract should be displayed on all of the account balance sections when connected to the account that owns the delegate contract.

I set up a delegate contract on Goerli for testing and I was able to vote:
image
image
image


rafinskipg (commented):
My question is , why do we have the boolean checkDelegate on getMKRVotingWeight?

PR #727, opened 2023-02-06
Add delegates endpoint pagination

hernandoagf:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/1825/add-pagination-to-delegates


modules/delegates/api/fetchDelegates.ts
```diff
@@ -302,3 +313,139 @@ export async function fetchDelegates(
 
   return delegatesResponse;
 }
+
+export async function fetchDelegatesPaginated(
+  queryParams: DelegatesAPIQueryParams
```
rafinskipg, line 318: do not send query params to the method, send "start" "limit" or "page" "pageSize" parameters

Query params are related to the API endpoint. This is a function that should not know about them

modules/delegates/api/fetchDelegates.ts
```diff
@@ -302,3 +313,139 @@ export async function fetchDelegates(
...
+      ? {
+          and: [
+            { voteDelegate: { notIn: ghVoteDelegateAddresses } },
+            { delegate: { notIn: migratedDelegateAddresses } }
+          ]
+        }
+      : null;
+
+  const delegatesQueryVariables = { ...queryParams };
+  if (delegatesQueryFilter) delegatesQueryVariables['filter'] = delegatesQueryFilter;
```
rafinskipg, line 352: do not use inline ifs

pages/api/delegates/delegates.ts
```diff
@@ -0,0 +1,57 @@
...
+    defaultValue: 'DESC',
+    validValues: ['ASC', 'DESC']
+  }) as string;
+
+  const status = validateQueryParam(req.query.status, 'string', {
+    defaultValue: 'ALL',
+    validValues: ['RECOGNIZED', 'SHADOW', 'ALL']
+  }) as string;
+
+  const validatedQueryParams = { network, first, after, includeExpired, orderBy, orderDirection, status };
```
rafinskipg, line 51: use the same standard of pagination that is found across the api (page / pageSize)

the start/ first are more related to the graphql implementation. The API should transform page/pageSize to whatever is needed by the graphql backend

modules/delegates/api/fetchDelegates.ts
hernandoagf: I should change the name of that function argument, it's not really the raw query params that the API endpoint is receiving, but the validated, filtered list of them, so I'm just passing to that function the params I need to return the right list of delegates

pages/api/delegates/delegates.ts
hernandoagf: Are you referring to only the names of the parameters or that we should change `after: Cursor` for `page: number`? The GraphQL API generates a random Cursor that is used for pagination, which would be a bit difficult to change to a page number, unless we stored the cursor somewhere after each GraphQL response

modules/delegates/api/fetchDelegates.ts
hernandoagf: I'll change these, I think I've used them in a couple of places

pages/api/delegates/delegates.ts
rafinskipg: oh, interesting. can you explain me what are the benefits of using the cursor versions increasing the start/limit values?

hernandoagf: Actually, I read about a lot of benefits of cursor pagination over offset pagination, but doing a deeper investigation seems like most of them don't really apply to us :/
- I don't think we can avoid the situation where the API endpoint would return some duplicate or missing entries as we're ordering the rows at the time of execution (based on MKR or delegator count), so if the delegations change, a delegate could be missed on one of the pages. That said, I think that's unlikely to happen to recognized delegates as all of them will be returned on the first or second page.
- Cursor pagination is claimed to be faster as the SQL query doesn't have to iterate over the full table, but just select the requested number of records after the provided unique identifier, but despite Postgraphile returning cursors on paginated queries, it is actually just running offset pagination in the background, so the performance improvements cannot be applied here.

I'll switch the API to use offset for pagination (which would allow us to use page numbers), it would involve adding a couple of extra lines of code but I think it'll be easier to interact with.

rafinskipg: cool, agree with the points, and the pagination with changing data is something we can not avoid, is what it is

modules/delegates/helpers/filterDelegates.ts
```diff
@@ -59,3 +61,41 @@ export function filterDelegates(
       })
   );
 }
+
+export function filterDelegateAddresses(
+  githubDelegates: DelegateRepoInformation[] | undefined,
+  allDelegateAddresses: AllDelegatesEntry[],
```
rafinskipg, line 67: wouldn't it be better to already have joined the data? 

I believe there's an entity called Delegate, (you can also use a partial delegate) instead of sending 2 different arrays. 

You would just need to filter over 1 entity.

rafinskipg: Also adding some unit tests for the filter function could be good

modules/delegates/helpers/getDelegatesCounts.ts
```diff
@@ -0,0 +1,56 @@
+import { SupportedNetworks } from 'modules/web3/constants/networks';
+import { AllDelegatesEntry, DelegateRepoInformation } from '../types';
+import { add } from 'date-fns';
+import { getNewOwnerFromPrevious } from 'modules/migration/delegateAddressLinks';
+
+export default function getDelegatesCounts(
+  githubDelegates: DelegateRepoInformation[] | undefined,
+  allDelegateAddresses: AllDelegatesEntry[],
```
rafinskipg, line 8: Same, joining delegates into one single entity would make things easier. You can also get the counts by filtering by the property "recognized"

pages/api/delegates/delegates.ts
```diff
@@ -0,0 +1,83 @@
+/*
+
```
rafinskipg, line 2: shouldn't we replace api/delegates/index.ts?

rafinskipg (changes requested):

pages/api/delegates/delegates.ts
hernandoagf: Ideally, yes. But I didn't want to break functionality for services relying on the current delegates API endpoint.

Maybe we can use api/delegates/index.ts for this one and rename the current one to api/delegates/legacy.ts while we give others time to adjust to the new workflow?

modules/delegates/types/delegate.d.ts
```diff
@@ -67,6 +67,40 @@ export type Delegate = {
   };
 };
 
+export type DelegatePaginated = {
+  name: string;
+  voteDelegateAddress: string;
```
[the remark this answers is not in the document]
hernandoagf: Good point. In this case, I'm adding some extra props but also removing some of them, like the list of delegations for each delegate. I can use Omit<Delegate, "mkrLockedDelegate"> for that :)

modules/delegates/helpers/filterDelegates.ts
hernandoagf: I'll be adding the unit tests tomorrow

tyler17 (commented):
I get this error when I hit the endpoint
Screen Shot 2023-02-15 at 4 37 33 PM
Is that because the necessary changes aren't on gov polling db yet?

modules/delegates/api/fetchDelegatesExecSupport.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  try {
+    const chainId = networkNameToChainId(network);
+
+    const data = await gqlRequest<Query>({ chainId, query: allDelegates });
+    const delegates = data.allDelegates.nodes;
+
+    const contracts = getContracts(chainId, undefined, undefined, true);
+    const delegatesExecSupport = await Promise.all(
+      delegates.map(async delegate => {
```
tyler17, line 35: You were saying it takes a while to fetch the exec support for all the delegates?  But you think this is better than only passing in the subset of delegates we care about? Alternatively, I wonder if it'd be fine to only have the UI show the exec support for recognized delegates on the main delegates page

pages/api/delegates/index.ts
```diff
@@ -7,25 +7,77 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 */
```
tyler17, line 7: There's a merge conflict in this file because Rafa added swagger api definitions to this function, so they should be moved to the legacy endpoint. And it would be nice to add swagger api definitions to this endpoint too so that people can see this endpoint on the api-docs page

modules/web3/constants/networks.ts
```diff
@@ -59,7 +59,7 @@ export const CHAIN_INFO: ChainInfo = {
     spockUrl: process.env.NODE_ENV === 'development' ? STAGING_MAINNET_SPOCK_URL : MAINNET_SPOCK_URL,
     rpcs: {
       [NodeProviders.INFURA]: `https://mainnet.infura.io/v3/${config.INFURA_KEY}`,
-      [NodeProviders.ALCHEMY]: `https://eth-mainnet.alchemyapi.io/v2/${config.ALCHEMY_KEY}`
+      [NodeProviders.ALCHEMY]: `https://eth-mainnet.g.alchemy.com/v2/${config.ALCHEMY_KEY}`
```
tyler17, line 62: is there a difference between these endpoints?

modules/delegates/helpers/filterDelegates.ts
```diff
@@ -59,3 +59,20 @@ export function filterDelegates(
...
 }
+
+export function filterDelegateAddresses(
+  allDelegatesWithNames: AllDelegatesEntryWithName[],
+  queryTags: string[] | null,
+  name: string | null
+): string[] {
+  const filteredDelegates =
+    !queryTags && !name
+      ? allDelegatesWithNames.filter(delegate => delegate.name)
```
tyler17, line 70: so this filters out shadow delegates?

modules/delegates/api/fetchDelegates.ts
```diff
@@ -302,3 +312,157 @@ export async function fetchDelegates(
...
+          ? { address: oldOwner, voteDelegateAddress: oldContractAddress }
+          : undefined,
+      next:
+        newOwner && newContractAddress
+          ? { address: newOwner, voteDelegateAddress: newContractAddress }
+          : undefined
+    };
+  });
+
+  console.log(allDelegatesWithNamesAndLinks.filter(d => d.name === 'Flip Flop Flap Delegate LLC'));
```
tyler17, line 365: can delete console.log

hernandoagf:
> 

@tyler17 I think so, I was taking a look at the logs on Vercel and I think these deployments are pointing to the production env of the gov-polling-db. Changes are currently live on staging

modules/comments/components/AccountComments.tsx
```diff
@@ -58,10 +58,10 @@ export default function AccountComments({ address }: { address: string }): React
           >
             Comments
           </Text>{' '}
-          {data.comments.map(comment => (
+          {data.comments.map((comment, i) => (
             <Box
               sx={{ borderBottom: '1px solid', borderColor: 'secondaryMuted', py: 4 }}
-              key={comment.address.address}
+              key={comment.address.address + '-' + i}
```
adamgoth, line 64: Do we have anything else unique on the comment object we can use as a key? Maybe the date? Using a mapping index as a key can lead to weird behavior in some situations (probably fine in this case since we're not adding/removing things, but still would be better to avoid it if possible).

modules/delegates/api/__tests__/filterDelegateAddresses.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+    expired: false,
+    name: 'coldiron.eth',
+    delegateType: DelegateTypeEnum.RECOGNIZED,
+    tags: ['growth', 'risk', 'security', 'decentralization'],
+    previous: undefined,
+    next: {
+      address: '0xe491d9ea59cae5dd84804b224535c64d5cb90c3a',
+      voteDelegateAddress: '0xb4b82978fce6d26a22dea7e653bb9ce8e14f8056'
+    }
+  },
```
adamgoth, line 28: Maybe we can move this to a json file so it could be used in other places as well. For example, we have a mock polls json at `modules/polling/api/mocks`

modules/delegates/components/DelegateOverviewCard.tsx
```diff
@@ -322,17 +337,19 @@ export function DelegateOverviewCard({ delegate }: PropTypes): React.ReactElemen
...
         />
       )}
       {showUndelegateModal && (
         <UndelegateModal
           delegate={delegate}
           isOpen={showUndelegateModal}
           onDismiss={() => setShowUndelegateModal(false)}
-          mutateTotalStaked={mutateTotalStaked}
+          mutateTotalStaked={mutateDelegateTotalMkr}
           mutateMKRDelegated={mutateMKRDelegated}
```
adamgoth, line 351: What is the difference between `mutateTotalStaked` and `mutateMKRDelegated`? They seem like they would serve the same purpose.

And do you know the reason why `refetchOnDelegation` is sometimes `true` and sometimes `false`?

modules/delegates/components/TopDelegates.tsx
```diff
@@ -182,9 +179,7 @@ export default function TopDelegates({
                       <Text as="p" variant="caps" sx={{ color: 'secondaryEmphasis' }}>
                         Delegators
                       </Text>
-                      <Text>
-                        {delegationHistory.filter(i => new BigNumber(i.lockAmount).gt(0)).length} addresses
-                      </Text>
+                      <Text>{delegate.delegatorCount} addresses</Text>
```
adamgoth, line 182: Much nicer 😌

tyler17 (dismissed):
This looks good as far as I can tell, I tried to look over everything.  It's a lot of changes though, we should definitely be sure to QA.

modules/address/components/Avatar.tsx
```diff
@@ -24,25 +24,30 @@ export function Avatar({ size, address, defaultComponent, style }: AvatarProps):
...
-      try {
-        const ensName = await library.lookupAddress(address);
-        if (ensName) {
-          const uri = await avt.getAvatar(ensName, {});
-          if (uri) {
-            setAvatarUri(uri);
+  useEffect(() => {
+    let mounted = true;
+
+    const fetchAvatarUri = async () => {
```
rafinskipg, line 30: i learned recently that there's a more optimal way to do this,

useSWR and cache the key to the address avatar, it will reduce the requests on the app considerably, 

Lets do this in a new PR @hernandoagf , once this is merged

modules/delegates/hooks/useDelegatedToWithDelegate.ts
```diff
@@ -0,0 +1,29 @@
...
+import { SupportedNetworks } from 'modules/web3/constants/networks';
+import { MKRDelegatedToWithDelegateAPIResponse } from 'pages/api/address/[address]/delegatedToWithDelegates';
+import useSWR from 'swr';
+
+export function useDelegatedToWithDelegates(
+  address: string | undefined,
+  network: SupportedNetworks
+): MKRDelegatedToWithDelegateAPIResponse | undefined {
+  const { data } = useSWR<MKRDelegatedToWithDelegateAPIResponse>(
+    address ? `/api/address/${address}/delegatedToWithDelegates?network=${network}` : null,
```
rafinskipg, line 19: this route name is a bit weird

pages/api/delegates/namesAndMetrics.ts
```diff
@@ -0,0 +1,109 @@
...
+ *      - voteDelegateAddress
+ *      - status
+ *      - blockTimestamp
+ *      - expirationDate
+ *      - expired
+ *      - isAboutToExpire
+ */
+
+export default withApiHandler(async (req: NextApiRequest, res: NextApiResponse<DelegateNameAndMetrics[]>) => {
+  const network = validateQueryParam(req.query.network, 'string', {
```
rafinskipg, line 101: Again, the URL does not need to specify all the returned fields.

rafinskipg (changes requested):
When we do a search by text, it wants us to click the icon, it should be able to trigger with the enter

image


Also it only filters the recognized ones.


I don't understand why we have a delegatesNamesAndMetrics and delegates distinction, i need your info on this , in order to be able to review the PR fully.

rafinskipg (approved):
In general "useDelegateInfo" (singular) should be enough to determine that is for a single delegate.

But i approve ✔️

modules/delegates/api/fetchDelegates.ts
```diff
@@ -302,3 +314,281 @@ export async function fetchDelegates(
...
+    )
+    .filter((delegate, i, arr) =>
+      recognizedOnly && !includeExpired ? arr.findIndex(d => d.name === delegate.name) === i : true
+    )
+    .sort((a, b) => new Date(a.blockTimestamp).getTime() - new Date(b.blockTimestamp).getTime())
+    .map(delegate => {
+      const githubDelegate = githubDelegates?.find(d => d.name === delegate.name);
+      return {
+        name: delegate.name || 'Shadow Delegate',
+        picture: githubDelegate?.picture,
```
rafinskipg, line 421: this code merging the delegates structure is repeated 3 times in this page, please, move it to another function and make the other functions reuse it, if possible

pages/api/delegates/legacy.ts
```diff
@@ -0,0 +1,204 @@
...
+*/
+
+import { NextApiRequest, NextApiResponse } from 'next';
+import { fetchDelegates } from 'modules/delegates/api/fetchDelegates';
+import { DelegatesAPIResponse } from 'modules/delegates/types';
+import { DEFAULT_NETWORK, SupportedNetworks } from 'modules/web3/constants/networks';
+import withApiHandler from 'modules/app/api/withApiHandler';
+import validateQueryParam from 'modules/app/api/validateQueryParam';
+
+/**
```
rafinskipg, line 16: do we want to keep this legacy thing?

if we want to version the API i suggest we add a v1/v2 folder or route, 

routes mapping can be defined in next.config, 

but i think we should check with whoever team is using our API endpoints, ,and notify them of the migration

if the data structure changed , if not, is okay

rafinskipg (changes requested):

rafinskipg (approved):

next.config.js
```diff
@@ -84,6 +84,10 @@ const moduleExports = {
       {
         source: '/delegates/:address',
         destination: '/address/:address'
+      },
+      {
+        source: '/api/delegates',
+        destination: '/api/delegates/v1'
```
rafinskipg, line 90: neat

PR #733, opened 2023-02-13
Fetch delegatedata aggregated

rafinskipg:
Fetch all the delegates information from the aggregated json


hernandoagf:
It all looks good to me. Tested it locally and it's returning all of the necessary data. It won't let me approve the PR as it is still a draft, though.

PR #748, opened 2023-03-28
Add vercel analytics

rafinskipg:
https://vercel.com/docs/concepts/analytics/audiences/quickstart

REMOVE all old analytics info


adamgoth (dismissed):

adamgoth (commented):
Is it possible we can also remove all mixpanel and analytics code that already exists?

adamgoth (approved):
Do you think we still need the cookies component functionality?

PR #749, opened 2023-03-31
Cvc filter

tyler17:


modules/delegates/types/cvc.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+SPDX-FileCopyrightText: © 2023 Dai Foundation <www.daifoundation.org>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+
+*/
+  
+export type CvcAndCount = {
+    cvc_name: string;
+    count: number;
```
tyler17, line 11: I copied this from the TagCount type, but we don't seem to use the count property as far as I can tell?

adamgoth (approved):
lgtm 💸

PR #759, opened 2023-04-07
Delegates list QA - Sorting filter and go back from detail screen

torhector2:
This PR tries to fix some issues discovered during QA

- Keep filters when you visit a delegate detail page and the press the "< Back to all delegates" button.
- Sort filter button shows right selected value after coming from delegate detail page. It was showing the filter without text.
- Change from "Delegate" to "Delegate MKR" in the delegate card.
- Apply reset to the sort button
- FIX: When you search by text (applying filters optionally), visit the profile and back the search term disappeared and the filter by text isn't applied.
- Perform request keeping active filters when user clears the name field.

### DIFF
--- a/modules/app/components/filters/SearchBar.tsx
+++ b/modules/app/components/filters/SearchBar.tsx
@@ -17,6 +17,7 @@ type Props = {
   value: string | null;
   placeholder?: string;
   withSearchButton?: boolean;
+  performSearchOnClear?: boolean;
 };
 
 export const SearchBar = ({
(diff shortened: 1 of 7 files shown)


hernandoagf (approved):
Looks good to me! I'll just add a small extra commit passing the `onVisitDelegate` function to the `DelegateAvatarName` component as you can also navigate to a delegate detail page by clicking on their names.

hernandoagf:
Added another commit fixing the delegate page lag when the delegates list becomes long

adamgoth (approved):
LGTM, nice work with all these 👷‍♀️

PR #764, opened 2023-04-12
landing page delegate fetching updates

adamgoth:


modules/app/components/InternalLink.tsx
```diff
@@ -33,7 +33,8 @@ export const InternalLink = ({
     scroll={scroll}
     title={title}
     sx={{ textDecoration: 'none', color: 'inherit', ...styles }}
+    passHref
   >
-    {children}
+    <a sx={{ textDecoration: 'none', color: 'inherit', ...styles }}>{children}</a>
```
rafinskipg, line 38: this sx property on a "non theme-ui" element causes the linter to fail, i tried solving it before but the only way i saw was either just using theme-ui elements or adding a tslint ignore on top of the file

rafinskipg (approved):

PR #767, opened 2023-04-17
Load more delegates

hernandoagf:
- Fix bug with pages metadata title
- Added functionality to the delegates page to avoid fetching all shadow delegates

### DIFF
--- a/pages/delegates/index.tsx
+++ b/pages/delegates/index.tsx
@@ -101,6 +101,12 @@ const Delegates = ({
     setFetchOnLoad(false);
   };
 
+  const handleLoadAllClick = () => {
+    setLoadAllDelegates(true);
+    setEndOfList(false);
+    setShouldLoadMore(true);
+  };
+
   const [loading, setLoading] = useState(fetchOnLoad);
   const [isRendering, setIsRendering] = useState(true);
   const [shouldLoadMore, setShouldLoadMore] = useState(false);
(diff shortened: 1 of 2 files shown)


tyler17 (commented):
Maybe we hide Load all delegates button after it's already been clicked? Just an idea
Screen Shot 2023-04-17 at 10 59 01 PM

tyler17 (approved):

hernandoagf:
Makes sense, I'll add that

pages/delegates/index.tsx
```diff
@@ -375,6 +395,24 @@ const Delegates = ({
                   ))}
                 </Stack>
               )}
+
+              {delegates.length && (
```
b-pmcg, line 399: haven't tested it, but I assume if delegates.length === 0 this will return false? Also do we need to check if delegates is undefined?

hernandoagf: My bad, I was going to change that line as that syntax is ambiguous. But yeah, as 0 is a falsy value that evaluates to false and that JSX is skipped.
Theoretically, I think delegates can't be undefined as the state variable is always set to an array, either empty or with values. What I'm seeing is that the page may break if the API call fails (delegates returned from the API would be undefined and when using the spread operator on undefined it throws an error).

I need to re-check if the API endpoint throws an error on failure or if it returns empty keys

rafinskipg (approved):

PR #775, opened 2023-04-25
Fix Tabs component routing in React 18

hernandoagf:
Found an error caused by the Tabs component affecting both local and production environments. Read these messages for context and references: https://discord.com/channels/836646301986848768/854643254741565450/1100528454421713026


tyler17 (approved):
Nice, confirmed it fixes the issue for me

PR #778, opened 2023-04-28
React hydration errors fix

hernandoagf:


modules/esm/components/ProgressRing.tsx
```diff
@@ -50,13 +48,7 @@ const ProgressRing = ({
...
         ></circle>
         <text x="50%" y="48%" textAnchor="middle" fill="#434358" fontSize="18px" dy=".3em">
-          {totalStaked ? (
-            `${formatValue(totalStaked, 'wad', 6)}      `
-          ) : (
-            <Box pl="14px" pr="14px">
-              <div ref={loader} />
-            </Box>
-          )}
+          {totalStaked ? `${formatValue(totalStaked, 'wad', 6)}      ` : '---      '}
```
b-pmcg, line 51: is this supposed to have that many spaces after the ---?

modules/home/components/ResourcesLanding/ResourcesLanding.tsx
```diff
@@ -6,29 +6,34 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
 import { ExternalLink } from 'modules/app/components/ExternalLink';
 import { fadeIn } from 'lib/keyframes';
 import { alpha } from '@theme-ui/color';
 
 const CategoryButton = ({ label, color, active, onClick }) => {
+  const { theme } = useThemeUI();
+  const [badgeSelectorColor, setBadgeSelectorColor] = useState('#000');
+
+  useEffect(() => {
+    setBadgeSelectorColor(theme.rawColors?.badgeSelector as string);
```
b-pmcg, line 21: if `theme.rawColors?.badgeSelector` is undefined maybe we should pass the default value in

modules/esm/components/ProgressRing.tsx
hernandoagf: Seems like the format is still correct without the extra spaces, I'll remove those

b-pmcg (approved):
🚀

tyler17:
Do you have an example exec where the statbox looked weird?

hernandoagf:
> Do you have an example exec where the statbox looked weird?

@tyler17 The UI is ok, but it would throw a React hydration error on the console because `div` tags cannot appear as children of `p` tags

tyler17:
Ah got it, thanks! All looks good to me then.

tyler17 (approved):

PR #785, opened 2023-05-12
Delegates endpoint counts fix

hernandoagf:
### What does this PR do?
It fixes the delegate and CVC counts, making them dynamic depending on the filters applied to the API request.
Now, when applying a CVC filter, the count of shadow delegates should be 0 and when filtering for only shadow delegates, all CVCs should display a count of 0.

Note that when select a CVC, the count of other CVCs remain the same. This is intended as a single delegate contract cant belong to more than 1 CVC.


modules/delegates/api/fetchDelegates.ts
```diff
@@ -456,10 +456,20 @@ export async function fetchDelegatesPaginated({
...
-  const { constitutionalDelegatesCount, shadowDelegatesCount, totalDelegatesCount } = getDelegatesCounts(
-    allDelegatesWithNamesAndLinks
+  const filteredDelegateAddresses = filterDelegateAddresses(
+    allDelegatesWithNamesAndLinks,
+    cvcs,
+    searchTerm,
+    delegateType
   );
+  const filteredDelegateEntries =
+    !searchTerm && !cvcs
```
rafinskipg, line 466: why dont you just refactor the filterDelegateAddresses to also return the data you need in fiteredDelegateEntries,

meaning, you change filterDelegateAddresses to a filterDelegates function, and then use the result

tyler17 (dismissed):
looks good to me and works as expected, thanks for updating 👍

b-pmcg (approved):
LGTM, builds fine. I just updated nvmrc so I could build it locally

tyler17:
Pretty sure this is just an issue with my local env for some reason because I've had it before, and I believe @hernandoagf mentioned it might be because of the cache, but I don't see the AVC filter on mainnet on this deployment, only goerli 🤷 I also have the issue on staging deployment, but not on production.  I think I previously had it on production before. Weird.

tyler17 (approved):

modules/delegates/helpers/filterDelegates.ts
```diff
@@ -6,77 +6,52 @@ SPDX-License-Identifier: AGPL-3.0-or-later
...
+    queryAvcs,
+    searchTerm,
+    type
   );
+  const filteredDelegateEntries =
+    !searchTerm && !queryAvcs
+      ? allDelegatesWithNames
+      : allDelegatesWithNames.filter(delegate => filteredDelegateAddresses.includes(delegate.voteDelegate));
+
+  return { alignedDelegatesAddresses, filteredDelegateAddresses, filteredDelegateEntries };
```
rafinskipg, line 34: i still feel that this is a bit weird, is a filter function but it returns an unfiltered result with the addresses. I'll approve it but in the future i would advice to maybe make this code refactored in a way that is more easy to understand what it does

rafinskipg (approved):

adamgoth (approved):

PR #786, opened 2023-05-17
Fix vote proxy contract gasless error

hernandoagf:
### What does this PR do?
It fixes the UI bug that allowed you to proceed to gasless voting without meeting the requirements.

### DIFF
--- a/modules/polling/components/review/ReviewBox.tsx
+++ b/modules/polling/components/review/ReviewBox.tsx
@@ -32,11 +32,15 @@ import EtherscanLink from 'modules/web3/components/EtherscanLink';
 
 export default function ReviewBox({
   account,
+  delegateAddress,
+  proxyAddress,
   activePollCount,
   activePollIds,
   ballotPollIds
 }: {
   account: string;
+  delegateAddress: string | undefined;
+  proxyAddress: string | undefined;
   activePollCount: number;
   activePollIds: number[];
   ballotPollIds: string[];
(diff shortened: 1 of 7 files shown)


tyler17:
Confirmed the UI fix looks good but by passing in the wallet address to the precheck we now have the same issue with delegate contracts, where the precheck is using the wallet address but the vote endpoint is using the delegate contract address when checking if it recently used the gasless service.

modules/polling/components/review/ReviewBox.tsx
```diff
@@ -59,7 +63,9 @@ export default function ReviewBox({
 
   const { data: precheckData } = useSWR(
     account && ballotPollIds && ballotPollIds.length > 0 && network
-      ? `/api/polling/precheck?network=${network}&voter=${account}&pollIds=${ballotPollIds.join(',')}`
+      ? `/api/polling/precheck?network=${network}&voter=${account}${
```
rafinskipg, line 66: i think it would be better if you just send the wallet address, and in the api you check if it has any proxy or delegateaddress instead of having to send all the info

rafinskipg (changes requested):

modules/polling/components/review/ReviewBox.tsx
tyler17: I was actually going to say I don't think there's any issue just using the wallet address when dealing with the cache check in the precheck and vote api.  The point of that check is to make sure someone can't vote in the same poll 100 times before their vote shows up in spock. It shouldn't matter if that address has a vote proxy or delegate contract.

hernandoagf: For the cache is fine to use the user's wallet address, but I'm not sure if we can use only that when fetching the voting history. I'll check that.

My initial approach was fetching the proxy contract and delegate contract on the precheck endpoint but it makes the user interaction slower

tyler17 (approved):
Nice this feels cleaner. I tested it out with regular wallet, delegate contract, and hot and cold addresses of a vote proxy, all looks good

rafinskipg (approved):

PR #787, opened 2023-05-22
Add walletconnect v2 support

hernandoagf:
### Link to Shortcut ticket:
https://app.shortcut.com/dux-makerdao/story/789/update-walletconnect-to-2-0

This is the permalink to test this branch https://governance-portal-v2-git-walletconnect-v2-dux-core-unit.vercel.app/
There's an allowlist for the WalletConnect Project ID that won't make it work on every URL. We can update the allowlist, though.


lib/config.ts
```diff
@@ -26,6 +26,7 @@ type SystemConfig = {
   DEFENDER_API_SECRET_MAINNET: string;
   ALCHEMY_ARBITRUM_KEY: string;
   ALCHEMY_ARBITRUM_TESTNET_KEY: string;
+  WALLETCONNECT_PROJECT_ID: string;
```
adamgoth, line 29: Looks like we still need to add a value for this in vercel for production. Just flagging so we don't forget.

modules/delegates/hooks/useDelegateCreate.ts
```diff
@@ -35,7 +38,15 @@ export const useDelegateCreate = (): LockResponse => {
   );
 
   const create = callbacks => {
-    const freeTxCreator = () => voteDelegateFactory.create();
+    if (!account || !provider) {
+      return;
+    }
+
+    const freeTxCreator = async () => {
```
adamgoth, line 45: I know you didn't name this function but I'm seeing it should actually be called `createTxCreator` instead of `lockTxCreator`. Might as well update in this PR.

Also noticing we define `LockResponse` as a type in many of these hooks files. We should probably give that a better name and export it from somewhere. Again, not something you wrote in this PR but maybe a good opportunity to update it.

adamgoth (commented):
LGTM. Left a few minor suggestions for the PR. 

I also did a quick test with a wallet connected with walletconnect and it seemed to work as expected, but I think once we have this on staging, we'll want to do a bit more extensive testing with sending transactions since this updates the code for sending transaction in general.

torhector2 (dismissed):
🚀 LGTM 👍

modules/polling/context/BallotContext.tsx
```diff
@@ -330,6 +330,7 @@ export const BallotProvider = ({ children }: PropTypes): React.ReactElement => {
             );
         },
         error: (txId, error) => {
+          console.log(error);
```
b-pmcg, line 333: did you mean to leave this log here?

modules/web3/helpers/sendTransaction.ts
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+import { providers } from 'ethers';
+
+export async function sendTransaction(
+  populatedTransaction: providers.TransactionRequest,
+  provider: providers.Web3Provider,
+  account: string
+): Promise<providers.TransactionResponse> {
+  const gasLimit = await provider.estimateGas(populatedTransaction);
```
b-pmcg, line 16: maybe we should wrap this in a try/catch and set a manual gas limit if it fails

b-pmcg (approved):
nice, LGTM let's just make sure to test all the contract calls on mainnet too

PR #807, opened 2023-07-19
Executive supporters batch requests + mobile delegate names

hernandoagf:
### What does this PR do?
- Fix an issue where executive supporters were not being fetched due to an Alchemy rate limitation for the batch RPC requests
- Split the aligned delegate names into 2 rows in mobile screens, that way the delegate name will always be visible regardless of the length of the AVC name

### DIFF
--- a/modules/executive/api/fetchExecutiveVoteTally.ts
+++ b/modules/executive/api/fetchExecutiveVoteTally.ts
@@ -40,7 +40,8 @@ export async function fetchExecutiveVoteTally(chief: Chief): Promise<any | null>
     }
   });
 
-  // Split voters array into chunks of 1000 addresses
+  // We need to split the voters array into chunks of 1000 addresses because our
+  // Alchemy key doesn't support batch requests larger than 1000 queries each
   const chunkSize = 1000;
   const voterChunks = Array.from({ length: Math.ceil(voters.length / chunkSize) }, (_, i) =>
     voters.slice(i * chunkSize, i * chunkSize + chunkSize)
--- a/modules/executive/api/fetchExecutiveVoteTally.ts
+++ b/modules/executive/api/fetchExecutiveVoteTally.ts
@@ -12,6 +12,14 @@ import { getChiefDeposits } from 'modules/web3/api/getChiefDeposits';
 import { getSlateAddresses } from '../helpers/getSlateAddresses';
 import { formatValue } from 'lib/string';
 import { paddedBytes32ToAddress } from 'lib/utils';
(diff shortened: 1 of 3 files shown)


modules/executive/api/fetchExecutiveVoteTally.ts
```diff
@@ -32,36 +40,48 @@ export async function fetchExecutiveVoteTally(chief: Chief): Promise<any | null>
...
 
-  const withDeposits = await Promise.all(
-    voters.map(voter =>
-      getChiefDeposits(voter, chief).then(deposits => ({
-        address: voter,
-        deposits
-      }))
-    )
+  // Split voters array into chunks of 1000 addresses
+  const chunkSize = 1000;
```
adamgoth, line 44: Could also be helpful for future maintainers to leave a short comment about it being set to 1000 due the batch request limits

adamgoth (approved):
Left some minor suggestions but nice work figuring out the issue

tyler17 (approved):

modules/executive/api/fetchExecutiveVoteTally.ts
```diff
@@ -32,36 +40,49 @@ export async function fetchExecutiveVoteTally(chief: Chief): Promise<any | null>
...
-        address: voter,
-        deposits
-      }))
-    )
+  // We need to split the voters array into chunks of 1000 addresses because our
+  // Alchemy key doesn't support batch requests larger than 1000 queries each
+  const chunkSize = 1000;
+  const voterChunks = Array.from({ length: Math.ceil(voters.length / chunkSize) }, (_, i) =>
+    voters.slice(i * chunkSize, i * chunkSize + chunkSize)
   );
```
generatorpoint, line 48: Nice work, one thing I'm curious about though is how the `JsonRpcBatchProvider` handles batching on it's end. [From what I understand](https://github.com/ethers-io/ethers.js/issues/62#issuecomment-878558531), the provider just continuously adds batched requests to it's internal cache and executes them with a 10ms delay. 

If that's the case I wonder how the chunking here is actually working. But maybe it's because each loop of our chunk creates a new `pendingBatch` in the provider as seen [here](https://github.com/ethers-io/ethers.js/commit/d55ab6d4e6025c484cc7e64486d927bd54a0772b#diff-a068075bdc828cc0bfd94be99cf940f44e7c9854919535948622196dfac968b8R42-R46).

In any case it does seem to be working, so that's great

generatorpoint (approved):
Seems to work well, nice fix

PR #815, opened 2023-07-28
Feat:  AVCs page

hernandoagf:

### DIFF
--- a/pages/api/avcs/index.ts
+++ b/pages/api/avcs/index.ts
@@ -1,0 +1,115 @@
+/*
+
+SPDX-FileCopyrightText: © 2023 Dai Foundation <www.daifoundation.org>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+
+*/
+
+import validateQueryParam from 'modules/app/api/validateQueryParam';
+import withApiHandler from 'modules/app/api/withApiHandler';
+import { fetchAvcs } from 'modules/avcs/api/fetchAvcs';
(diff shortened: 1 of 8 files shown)


adamgoth (changes requested):
Nice work!

Overall looking good but I noticed a couple minors things

* the description doesn't render correctly
Screenshot 2023-08-08 at 11 45 40 AM

* we should hide the "View AVC details" button until the detail pages are ready to ship

* need to add AVCs to the mobile menu
Screenshot 2023-08-08 at 11 48 31 AM

* could use a little bit of love on mobile, although maybe not an issue if the AVC details button is removed
Screenshot 2023-08-08 at 11 48 40 AM

pages/api/avcs/index.ts
```diff
@@ -0,0 +1,115 @@
...
+import { DEFAULT_NETWORK, SupportedNetworks } from 'modules/web3/constants/networks';
+import { NextApiRequest, NextApiResponse } from 'next';
+
+/**
+ * @swagger
+ * paths:
+ *  /api/avcs:
+ *    get:
+ *      tags:
+ *        - "AVCs"
```
adamgoth, line 23: I'm not sure why but I don't see this show up on the API docs page as it's own section. I thought adding the tag of "AVCs" would have done that.

adamgoth (approved):

pages/api/avcs/index.ts
adamgoth: Apparently it was an issue with the last line where `$ref` needed to be on a new line. Fixed now.

PR #838, opened 2023-10-03
Move CSP to headers

adamgoth:


hernandoagf (approved):
LGTM, I see the CSP in the response headers when accessing the Gov Portal now.
I see CORS errors for WalletConnect on the deployed site but I think that's because the preview deployments are not on the list of allowed origins in the WC project. We can check if it works well in staging

PR #858, opened 2024-03-26
Provider errors

tyler17:
- make quorum 1, and add priorities to the providers.
- fix a couple calls that pass in a rpcUrl instead of a network
- decrease chunk size to 900
- run prettier and build


modules/web3/helpers/getDefaultProvider.ts
```diff
@@ -13,16 +13,19 @@ import { ethers, providers } from 'ethers';
...
+    const alchemyProvider = new ethers.providers.AlchemyProvider(network, config.ALCHEMY_KEY);
+    const pocketProvider = new ethers.providers.PocketProvider(network, config.POCKET_KEY);
+    const etherscanProvider = new ethers.providers.EtherscanProvider(network, config.ETHERSCAN_KEY);
+  
+    const provider = new ethers.providers.FallbackProvider([
+      { provider: infuraProvider, priority: 2 },
+      { provider: alchemyProvider, priority: 1 },
+      { provider: pocketProvider, priority: 3 },
+      { provider: etherscanProvider, priority: 3 }
+    ], 1); //Quorum of 1
```
generatorpoint, line 28: cool, so this will always try alchemy first, and if it is successful then we're good, right?

generatorpoint (approved):
awesome! I'm seeing the missing exec on the landing page and the MKR support on the executive page.

adamgoth (approved):

PR #867, opened 2024-04-16
Modify middleware to not return request headers in response object

hernandoagf:
### What does this PR do?
- Modify middleware to only return request headers in the `NextResponse.next.request` object
- Remove request headers from the response headers, and only set the CSP in the response headers


b-pmcg (approved):
great, nice find

PR #876, opened 2024-04-24
more tenderly updates

adamgoth:


modules/avcs/api/getAvcsRepositoryInfo.ts
```diff
@@ -22,6 +22,7 @@ export function getAvcsRepositoryInformation(network: SupportedNetworks): Reposi
     page: 'delegates/templates'
   };
 
-  const avcsRepositoryInfo = network === SupportedNetworks.MAINNET ? repoMainnet : repoTest;
+  const avcsRepositoryInfo =
+    network === SupportedNetworks.MAINNET || network === SupportedNetworks.TENDERLY ? repoMainnet : repoTest;
```
hernandoagf, line 26: If the plan is to only support Mainnet and Tenderly I think this can just point to the mainnet repo (no need for test repo). Same for the delegates repo in the following file

hernandoagf (approved):
LGTM, just a comment on the repos

tyler17 (approved):

PR #878, opened 2024-04-26
Generate new network sdk

adamgoth:
Add arb sepolia to sdk generator


tyler17 (approved):
Good catch 👍 Is it necessary to copy the abis over to a tenderly folder? Seems we didn't have to do it for the goerlifork hardhat network. But no harm in having it, will merge

PR #879, opened 2024-04-26
Remove goerli

adamgoth:
Removes all references to goerli


tyler17 (approved):
Looks good 👍 , confirmed tenderly still works for me. Will merge

PR #890, opened 2024-05-26
add polling e2e

tyler17:
- add a test to poll vote with legacy system

Tried to add a test for gasless voting, but it seems that our mock connector doesn't support eth_signTypedData_v4. We create the mock connector by extending ether's Eip1193Bridge, which doesn't support sign typed data. So it would be a bit of to update it to handle it I think, I left that test skipped.


adamgoth (approved):
You mentioned gasless not working but I see a couple things commented out in the `esmodule` test. Was this intentional?

PR #901, opened 2024-07-24
Liquidity fixes

tyler17:
Hardcode two Uniswap V3 pools and check their MKR balance (The next largest v3 pool only has 3 MKR in it so I think it's fine to hardcode the two popular ones. We're doing the same for uniswap V2).

Update the balancer subgraph url, and also add support for balancer V2.


b-pmcg (approved):
great! nice to have this back

adamgoth (approved):

PR #906, opened 2024-08-05
Tenderly updates

tyler17:
skip mock connector code if USE_MOCK_WALLET isn't true or if NODE ENV is production


b-pmcg (approved):
great! Errors gone

PR #918, opened 2024-09-26
Update migration page with v2 delegate requirements

torhector2:
Update migration page with v2 delegate requirements

TODOs:

- [x]  maybe delegator page needs to show migrated v2 delegates ???
- [x] ~Show v1 and v2 visual indications on delegates pages and list items ???~ done in separate pr

### DIFF
--- a/modules/migration/components/MigrationBanner.tsx
+++ b/modules/migration/components/MigrationBanner.tsx
@@ -22,13 +22,13 @@ export function MigrationBanner(): React.ReactElement | null {
     isDelegateContractExpired,
     isDelegateContractExpiring,
     isShadowDelegate,
-    isDelegateContractV1,
+    isDelegateV1Contract,
     isDelegatedToV1Contract
   } = useMigrationStatus();
   const showDelegationMigrationBanner =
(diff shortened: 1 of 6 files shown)


hernandoagf (approved):
Looking good so far!

modules/migration/components/MigrationBanner.tsx
```diff
@@ -21,19 +21,25 @@ export function MigrationBanner(): React.ReactElement | null {
     isDelegatedToExpiredContract,
     isDelegateContractExpired,
     isDelegateContractExpiring,
-    isShadowDelegate
+    isShadowDelegate,
+    isDelegateContractV1,
+    isDelegatedToV1Contract
```
adamgoth, line 26: Minor thing but we make name consistent here, `V1Contract`. They are currently flipped around in each variable name.

adamgoth (approved):
Looking good so far. Left one small suggestion.

modules/migration/components/MigrationBanner.tsx
torhector2: good catch

PR #936, opened 2024-10-08
Update MKR delegated by user

adamgoth:
* Use subgraph delegations to get total delegated by an address, so that lockstake delegated MKR is included
* Put "V2" on contract overview, rather than "No expiration"


tyler17 (approved):
Here's a PR to fix the failing e2e: https://github.com/makerdao/governance-portal-v2/pull/938
Otherwise looks good 👍

modules/delegates/components/modals/UndelegateModal.tsx
```diff
@@ -44,7 +44,7 @@ export const UndelegateModal = ({
   const voteDelegateAddress = delegate.voteDelegateAddress;
   const [mkrToWithdraw, setMkrToWithdraw] = useState(BigNumber.from(0));
 
-  const { data: mkrStaked } = useMkrDelegated(account, voteDelegateAddress);
+  const { data: mkrStaked } = useMkrDelegatedByUser(account, voteDelegateAddress);
```
torhector2, line 47: Can the user undelegate the MKR delegated outside the gov portal?

torhector2 (approved):

modules/delegates/components/modals/UndelegateModal.tsx
adamgoth: Good catch, will update this in another PR to only show the amount delegated directly, and add to the disclaimer info about the seal module

PR #938, opened 2024-10-09
fetch delegation amount from on chain if using a mock wallet, or the …

tyler17:
…subgraph call errors


torhector2 (approved):

modules/mkr/hooks/useMkrDelegatedByUser.ts
```diff
@@ -34,9 +38,26 @@ export const useMkrDelegatedByUser = (
...
+      abi,
+      chainId,
+      provider,
+      account,
+      true
+    );
+  
+    return contract.stake(userAddress as string);
+  };
+
```
adamgoth, line 54: Won't this only return the MKR delegated directly by the EOA, and not the MKR delegated via the lockstake position? That's why we had to make the switch to the subgraph.

adamgoth (approved):

PR #939, opened 2024-10-09
Show correct delegate amounts in undelegate modal

tyler17:
Also shorten (Seal Module) to (Seal) as it was too much text on mobile

Screenshot 2024-10-09 at 12 23 43 PM


adamgoth (approved):
Nice, good call to separate the amounts by delegation type in the hook

PR #966, opened 2024-12-06
Batch cached execs fetches and deletions

hernandoagf:
### What does this PR do?
- Replace the use of `redis.keys()` for `redis.scan()` which is optimized for production
- Batch exec proposals scans to 100 at a time
- Batch exec proposals deletions so they are deleted in groups of 100 instead of deleting them one by one


b-pmcg (approved):
nice work

tyler17 (approved):

PR #991, opened 2025-03-20
Add a unit test for fetchTally with subgraph

tyler17:
Adds a unit test for fetchPollTally (mocking the subgraph calls), and confirms that the poll tally is correct. Only tests a couple "majority" polls, but we already have tests for other types of polls using the spock based function.


tyler17:
to be replaced by another PR

PR #994, opened 2025-03-20
Legacy app: Disable write contract interactions

hernandoagf:
TODO:
- [ ] set the `READ_ONLY` env var to `true` in the deployment to disable all contract interactions except for undelegation and chief free

### What does this PR do?
- Add a `READ_ONLY` env var that can be used to determine if contract interactions will be disabled
- Disable all write contract interactions in the web app except for chief free (withdraw MKR), old chief free, delegate free (undelegating) and IOU tokens approval when the `READ_ONLY` env var is set to `true`
- Disable some buttons based on the `READ_ONLY` env var to avoid triggering modals with dead ends

List of interactions disabled:
- Poll creation
- Poll vote (legacy and gasless)
- Token approvals for DAI and MKR (still allowed for IOU and IOU_OLD as those are required for withdrawing)
- Chief lock - MKR deposit (EOA and proxy contract)
- Executive vote (EOA, proxy contract and delegate)
- Delegate contract creation
- Delegate lock
- ESM burn and ESM shutdown

### Steps for testing:
Attempt to perform any of the write contract interactions specified in the disabled list above, you shouldn't be able to do so

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -59,5 +59,5 @@ export const config: SystemConfig = {
   TENDERLY_RPC_KEY: process.env.NEXT_PUBLIC_TENDERLY_RPC_KEY || '',
   USE_MOCK_WALLET: process.env.NEXT_PUBLIC_USE_MOCK_WALLET || '',
   SUBGRAPH_API_KEY: process.env.NEXT_PUBLIC_SUBGRAPH_API_KEY || '',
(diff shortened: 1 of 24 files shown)


tyler17 (approved):
Looks good to me! Are we planning to have a banner or something that states that only withdrawal transactions are supported?

b-pmcg (approved):

PR #995, opened 2025-03-20
Exec supporters subgraph (both legacy and future app?)

tyler17:
Update the api/executive/supporters endpoint to use the subgraph instead of making a ton of rpc calls.

We'd need to deploy the subgraph update to production before deploying this.


hernandoagf (approved):
Nice, the list of supporters loads so much faster now. Also, I can't remember the last time I saw all these green lol
image

b-pmcg (approved):
a thing of beauty

PR #996, opened 2025-03-21
V3: Update theme and colors

hernandoagf:
### What does this PR do?
Update theme and colors to follow the Sky palette, using purple as the primary color


torhector2 (approved):

torhector2:
On big screens that submenu shows up in the middle of the page, which is okay and similar to the one in prod but I think it shouldn't have full width. I add an screenshot of the current version to compare.

image

image

hernandoagf:
> On big screens that submenu shows up in the middle of the page, which is okay and similar to the one in prod but I think it shouldn't have full width. I add an screenshot of the current version to compare.
> 
> image
> 
> image

You're right, I just reverted that

PR #1012, opened 2025-03-28
Batch RPC requests

hernandoagf:
### What does this PR do?
- Add the `multcall3` contract from mainnet to the Tenderly chain config
- Batch some contract reads into a single multicall read to reduce RPC requests
- Add the batch option to the Viem HTTP transports with a wait time of 500ms


b-pmcg (approved):

tyler17 (approved):
Nice!

b-pmcg (approved):

PR #1021, opened 2025-04-01
Make the enabled parameter of contract writes more explicit

hernandoagf:


b-pmcg (approved):
nice, hopefully reduces some RPC calls

PR #1046, opened 2025-04-09
Arbitrum poll create

tyler17:


pages/polling/create.tsx
```diff
@@ -58,6 +58,31 @@ const CreateText = ({ children }) => {
     </Box>
   );
 };
+
+const NetworkSwitchModal = () => {
```
hernandoagf, line 62: This modal blocks the app if the user is not connected to Arbitrum. It doesn't allow to switch networks, but if you move to a different page and switch there, you get an error. Maybe we can add a button to trigger Wagmi's `switchChain(arbitrum)` from this modal

hernandoagf (approved):
LGTM, but I think we should update the behavior of the modal in the create page

torhector2 (approved):

PR #1071, opened 2025-04-21
Update queries to work with the new subgraph entities

tyler17:
Only fetch v3 delegates and v2 chief events

Don't show executive mainnet data on tenderly


hernandoagf (approved):

hernandoagf (approved):
LGTM!

PR #1073, opened 2025-04-22
Adds BA labs USDS/DAI info endpoint

adamgoth:
### What does this PR do?

- Adds BA labs USDS/DAI info endpoint, replaces DAI supply with USDS
- Replace Dai Savings Rate for Sky Savings Rate data point


hernandoagf (approved):
Looking good so far

tyler17 (approved):

PR #1074, opened 2025-04-22
Update liquidity sidebar to also account for SKY

hernandoagf:
TODO:
- [ ] Do we need to fetch from any other sources? The remaining sources in the sidebar don't seem to have SKY liquidity (Aave, Balancer, Sushi, Compound)

### What does this PR do?
- Add the Uniswap pools with the most SKY liquidity
- Include SKY liquidity sources in the sidebar
- Redenominate total in terms of SKY as that is the token used in the app now

image


tyler17 (approved):
Looks great

PR #1078, opened 2025-04-24
Add deposit and withdraw buttons to ballot card

hernandoagf:
### What does this PR do?
It adds the ability to deposit and withdraw SKY into the chief from the ballot card in the polling page, this gives users the ability to manage their poll voting weight from the polling page directly since this voting weight will only count chief balance


modules/executive/components/ProgressBar.tsx
```diff
@@ -30,7 +30,7 @@ function CircleNumber({ selected = false, num }: { selected?: boolean; num: numb
 }
 
 export default function ProgressBar({ step }: { step: number }): JSX.Element {
-  const stepTexts = ['Withdraw', 'Deposit', 'Vote'];
```
tyler17, line 33: Maybe we can just remove this Progress bar entirely? I think it's a bit redundant with the vote address(0) flow on the account page, and looks kinda bad because it will flash before it realizes Ive already voted.

tyler17 (approved):
Looks good! Just a thought on the progress bar

b-pmcg (approved):

modules/executive/components/ProgressBar.tsx
tyler17: Small PR for this: <https://github.com/makerdao/governance-portal-v2/pull/1080>

PR #1081, opened 2025-04-29
unit test for voting address 0

sugh01:
- Add test for votedForAddressZero state and UI
- Add test for isChiefLive state and UI
- Add tests for delegate contract address display


b-pmcg (approved):
Nice, great work on this! Keep in mind some of the exact copy will change prior to launch, we'll make sure to update it here when it does.

PR #1115, opened 2025-05-13
Add mainnet staging metadata

tyler17:
When NEXT_PUBLIC_VERCEL_ENV = "development" we show the forked metadata instead of the makerdao metadata.

Alternatively, we could add a different environment variable that we use for this instead, if we want to default to makerdao metadata on development. What do we think?


adamgoth (approved):
I think this works. We can show the `main` branch metadata from the `jetstreamgg` repo in staging mainnet and `testnet` in staging testnet.

b-pmcg (approved):
I think this is a good approach, I notice the polling test is failing in other branches so I don't believe it was introduced here

PR #1191, opened 2025-07-14
show details for sky polls

b-pmcg:


tyler17 (dismissed):

torhector2 (approved):
LGTM! I've followed the testing steps and it works as expected.

One minor improvement would be to display the names when possible, the same way they appear in the sky portal. The two places I noticed this are the voting list and the weight voting circles.

PR #1256, opened 2026-07-28
fix e2e Tenderly transactions

0x2tester:
### What does this PR do?
Fixes the e2e suite, which was failing because Tenderly changed the RPC hostname for forked VNets. New forks return https://virtual.mainnet.eu.rpc.tenderly.co/… instead of https://virtual.mainnet.rpc.tenderly.co/…, so the page.route() pattern in playwright/forkVnet.ts silently stopped matching. That route is what raises the gas limit on eth_sendTransaction/eth_call, so every transaction-based test lost its gas bump.

### Steps for testing:

1. pnpm playwright install
2. Ensure .env has a valid TENDERLY_API_KEY and a valid GITHUB_TOKEN — an expired token 401s and the executive/delegate/poll lists render empty, which fails those specs for unrelated reasons.
3. pnpm e2e wallet.spec
4. pnpm e2e legacy-executives.spec / delegates.spec / legacy-polling.spec — these are the ones the gas-limit route actually gates.


playwright/forkVnet.ts
```diff
@@ -12,6 +12,34 @@ import { mockRpcCalls } from './mock-rpc-call';
 dotenv.config();
 
 const displayName = process.env.CI ? 'ci-tests-testnet' : 'local-tests-testnet';
+// id of the `e2e-testing-apr-15-fork_apr_29` vnet, used as the source of every test fork
+const DEFAULT_FORK_VNET_ID = 'bf27af19-1335-404a-9257-18affa774f5a';
```
fenris85: please move this value to env var

fenris85 (dismissed):

fenris85 (approved):

